PR #1, opened 2020-04-23
Problem: pulp_rpm 3.3.0 cannot install

mikedep333:
on the pulp container images because Fedora's createrepo_c
is too old

Solution: Temporarily replace the createrepo_c RPM with its
build-dependencies, and let createrepo_c get installed &
built from PyPI.

re: #6523

[noissue]


mikedep333:
pulpcore.git (pulp-operator) change: https://github.com/pulp/pulpcore/pull/658

PR #6, opened 2020-05-27
Prepare ssl keys for token_auth

mdellweg:
re #6782
https://pulp.plan.io/issues/6782

### DIFF
--- a/assets/pulpcore-api.prep
+++ b/assets/pulpcore-api.prep
@@ -1,0 +1,12 @@
+#!/usr/bin/execlineb -P
+
+if -n {
+  ls /var/lib/pulp/cert
+}
+foreground {
+  mkdir -p /var/lib/pulp/cert
+}
+foreground {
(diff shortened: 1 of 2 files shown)


assets/pulpcore-api.prep
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/execlineb -P
+
+foreground {
+  mkdir -p /var/lib/pulp/tmp
```
dkliban, line 4: Let's rename this directory to /var/lib/pulp/cert/

assets/pulpcore-api.prep
```diff
@@ -0,0 +1,9 @@
...
+}
+foreground {
+  openssl ecparam -genkey -name prime256v1 -noout -out /var/lib/pulp/tmp/private.pem
```
dkliban, line 7: Run this only when the file doesn't already exist.

mdellweg: Not so easy if you need to learn execline first.

PR #13, opened 2020-06-10
Change the way the version args work

daviddavis:
Allow users to specify an operator and default to nothing when building
pulp-oci-images container.

fixes #6955

### DIFF
--- a/.github/workflows/pr_build_images.yaml
+++ b/.github/workflows/pr_build_images.yaml
@@ -14,3 +14,14 @@ jobs:
           docker build --file pulp_ci/Containerfile --tag pulp/pulp-ci:latest .
           docker build --file pulp_fedora31/Containerfile --tag pulp/pulp-fedora31:latest .
           docker build --file pulp_galaxy_ng/Containerfile --tag pulp/pulp-galaxy-ng:latest .
+      - name: Test the images
+        run: |
+          mkdir settings
+          echo "CONTENT_ORIGIN='http://$(hostname):8080'" >> settings/settings.py
+          docker run --detach \
+                     --publish 8080:80 \
+                     --volume "/$(pwd)/settings:/etc/pulp:Z" \
+                     --device /dev/fuse \
+                     pulp/pulp-fedora31:latest
+          sleep 30
+          curl --fail localhost:8080/pulp/api/v3/status/
--- a/README.md
+++ b/README.md
@@ -12,3 +12,13 @@ $ <docker build | buildah bud> --file pulp_ci/Containerfile --tag pulp/pulp-ci:l
 $ <docker build | buildah bud> --file pulp_fedora31/Containerfile --tag pulp/pulp-fedora31:latest .
 $ <docker build | buildah bud> --file pulp_galaxy_ng/Containerfile --tag pulp/pulp-galaxy-ng:latest .
 ```
+
+## Specifying versions
+
+By default, containers get built using the latest version of each Pulp component. If you want to
+specify a version of a particular component, you can do so with args:
+
+```bash
+$ <docker build | buildah bud> --build_arg PULPCORE_VERSION="==3.5.0" --file pulp_fedora31/Containerfile
+$ <docker build | buildah bud> --build_arg PULP_FILE_VERSION=">=1.0.0" --file pulp_fedora31/Containerfile
+```
--- a/pulp_fedora31/Containerfile
+++ b/pulp_fedora31/Containerfile
@@ -1,20 +1,20 @@
 FROM pulp/pulp-ci:latest
 
-ARG PULPCORE_VERSION=3.6.0
-ARG PULP_FILE_VERSION=1.2.0
-ARG PULP_CONTAINER_VERSION=2.0.0
-ARG PULP_ANSIBLE_VERSION=0.2.0
-ARG PULP_RPM_VERSION=3.6.1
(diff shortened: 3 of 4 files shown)


daviddavis:
Sorry it took me so long to address this PR.

@fao89 I added some docs. 

@pieterlexis great points. 

> I like that a user can specify the operator, although I do consider the audience for using operators quite small. This 'advanced' usage of pip would be better done in another Containerfile using this image as a base.

I can drop the operators if that's to advanced for most people.

> It would be good to have some sane defaults in there if you're planning to publish images automatically. With this change, it will depend on pypi what versions would be included in the image. This means that building an image without setting the versions explicitly in the build command makes the resulting images (potentially) inconsistent.
>
> To bring this point home, the official nginx image has the version hard-coded as an ENV and I've seen this pattern in more Dockerfiles.

So part of the problem motivating this change is that these versions aren't getting updated so like right now [they're very outdated](https://github.com/pulp/pulp-oci-images/blob/126195f3ef4da700508e493c4034cecad36ecbc8/pulp_fedora31/Containerfile#L3-L8). I'm open to other ideas about how we keep these versions up to date but until then I think it would be better to give people the latest versions of packages from pypi rather than old releases.

pulp_fedora31/Containerfile
```diff
@@ -1,20 +1,20 @@
...
-ARG PULP_CONTAINER_VERSION=2.0.0b2
-ARG PULP_ANSIBLE_VERSION=0.2.0b13
-ARG PULP_RPM_VERSION=3.4.1
-ARG PULP_MAVEN_VERSION=0.1.0
+ARG PULPCORE_VERSION=""
+ARG PULP_FILE_VERSION=""
+ARG PULP_CONTAINER_VERSION=""
+ARG PULP_ANSIBLE_VERSION=""
+ARG PULP_RPM_VERSION=""
+ARG PULP_MAVEN_VERSION=""
```
fao89, line 8: I think we can merge it if we declare default values

daviddavis: So the point is not to have defaults but automatically use the latest versions from pypi.

.github/workflows/pr_build_images.yaml
```diff
@@ -14,3 +14,19 @@ jobs:
...
+          echo "CONTENT_ORIGIN='http://$(hostname):8080'
+                ANSIBLE_API_HOSTNAME='http://$(hostname):8080'
+                ANSIBLE_CONTENT_HOSTNAME='http://$(hostname):8080/pulp/content'
+                TOKEN_AUTH_DISABLED=True" >> settings/settings.py
+          docker run --detach pulp/pulp-fedora31:latest \
+                     --publish 8080:80 \
+                     --volume ./settings:/etc/pulp:Z \
+                     --volume ./pulp_storage:/var/lib/pulp:Z \
+                     --volume ./pgsql:/var/lib/pgsql:Z \
+                     --volume ./containers:/var/lib/containers:Z \
```
mdellweg, line 29: In my experience you only really need the settings volume. But maybe we want to test the way we document it...

.github/workflows/pr_build_images.yaml
```diff
@@ -14,3 +14,19 @@ jobs:
...
+                     --device /dev/fuse \
+                     pulp/pulp-fedora31
+          curl --fail localhost:80/pulp/api/v3/status/
```
mdellweg, line 32: You probably need to wait for a few seconds here.

README.md
```diff
@@ -12,3 +12,13 @@ $ <docker build | buildah bud> --file pulp_ci/Containerfile --tag pulp/pulp-ci:l
...
 $ <docker build | buildah bud> --file pulp_galaxy_ng/Containerfile --tag pulp/pulp-galaxy-ng:latest .
 ```
+
+## Specifying versions
+
+By default, containers get built using the latest version of each Pulp component. If you want to
+specify a version of a particular component, you can do so with args:
+
+```bash
+$ <docker build | buildah bud> --build_arg PULPCORE_VERSION="==3.5.0" --file pulp_fedora31/Containerfile
```
mdellweg, line 22: Is this collecting compatible versions of all the plugins in that case, or do you actually need to specify them all at this point?

mdellweg: BTW: Thank you for adding explicit docs around this.

daviddavis: I tested it out and ran `pip install --use-feature=2020-resolver pulpcore==3.3.0 pulp_file` which pulled down `pulp-file-0.3.0` so I think pip will handle it.

.github/workflows/pr_build_images.yaml
```diff
@@ -14,3 +14,14 @@ jobs:
...
+        run: |
+          mkdir settings
+          echo "CONTENT_ORIGIN='http://$(hostname):8080'" >> settings/settings.py
+          docker run --detach \
+                     --publish 8080:80 \
+                     --volume "/$(pwd)/settings:/etc/pulp:Z" \
+                     --device /dev/fuse \
+                     pulp/pulp-fedora31:latest
+          sleep 30
+          curl --fail localhost:8080/pulp/api/v3/status/
```
mdellweg, line 27: I think, i just saw a 503 bad gateway slipping through this command. We may need to actually read from the output.

mdellweg (approved):

pieterlexis:
The docs make things clear for the users. I think this is good to go indeed :)

PR #16, opened 2020-06-22
Ci pypi mirror

mdellweg:


dkliban:
Let's revisit this when we are ready to work on using this image as a PyPI mirror in our CI.

PR #17, opened 2020-06-22
Add Containerfile for Galaxy

mdellweg:
fixes #6713
https://pulp.plan.io/issues/6713

### DIFF
--- a/.github/workflows/pr_build_images.yaml
+++ b/.github/workflows/pr_build_images.yaml
@@ -10,5 +10,7 @@ jobs:
       - uses: actions/checkout@v2
       - name: Build the images
         run: |
-          docker build . --file Containerfile --target ci-base --tag pulp/pulp-ci:latest
-          docker build . --file Containerfile --tag pulp/pulp-fedora31:latest
+          wget https://github.com/just-containers/s6-overlay/releases/download/v1.22.1.0/s6-overlay-amd64.tar.gz
+          docker build --file pulp_ci/Containerfile --tag pulp/pulp-ci:latest .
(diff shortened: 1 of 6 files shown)


mdellweg:
With this new Containerfile i was able to build and run a container that provides the galaxy login screen. @chouseknecht

How do we want to deliver this? (In a separate image, or always installed in the common image?) @bmbouter, @dkliban 

Commands to reproduce:
```
buildah bud -f galaxy_ng/Containerfile -t pulp_galaxy_ng .
podman run --publish 8080:80 --device /dev/fuse --name pulp --volume ./settings:/etc/pulp pulp_galaxy_ng
```

chouseknecht:
@mdellweg having a separate image for galaxy_ng would be pretty sweet!

bmbouter:
Thanks @chouseknecht 

@mdellweg so that would be a container that ships: galaxy_ng, pulp_ansible, and pulp_contianer. Both container and ansible are dependencies of galaxy_ng. Would you be able to create that? How hard is it to auto-build-ship one of those too?

mdellweg:
This is exectly what the image contains. Buildshipping it is as hard as finding a proper name for it. Any suggests?

sivel:
I've tested this by switching to the `pulp-galaxy-ng` image in the ansible/ansible integration tests and it is passing.  See https://github.com/ansible/ansible/pull/70303

sivel:
I think we are all good with `pulp/pulp-galaxy-ng`.

bmbouter:
I agree, let's find a way to merge this. @mdellweg what do we need to do to merge this? I see one of the quality checks is still failing.

mdellweg:
I think, the build failure is due to pip19, so needs a rebase.

PR #18, opened 2020-06-24
Fix build error by using pip 20

daviddavis:
fixes #7042


daviddavis:
@mdellweg easy_install didn't seem to work so I had pip upgrade itself. Can you re-review?

PR #26, opened 2020-08-28
Update plugins in pulp_fedora31 container

mdellweg:


mdellweg:
I do not see this blocking your approach.
But it replaces the currently broken images on dockerhub. Something we should do rather soon.
Also using `--use-feature=2020-resolver` might be a good idea anyway.

daviddavis:
I think that #13 should also resolve the problem and when we discussed it at open floor earlier in the week, most people were in favor of the approach of not pinning pulpcore and pulp plugins.

mdellweg:
I am not sure, #13 alone solves that specific issue with attr. But again, the new depsolver can surely be combined with unpinning the versions.
On the other hand we might need some test to call the status endpoint of the freshly built image to avoid publishing one that is unable to even start the services.

That said, i believe this pr is not changing how we do things now, but only fixing that broken image already on dockerhub and in no way invalidating any discussion on #13 (it may need a rebase...).

daviddavis:
I don't want to block this. I can test my PR later and confirm it works.

PR #35, opened 2020-11-05
Adds pulp_python==3.0.0b12 to the container

dkliban:
[noissue]

### DIFF
--- a/pulp_fedora31/Containerfile
+++ b/pulp_fedora31/Containerfile
@@ -7,6 +7,7 @@ ARG PULP_ANSIBLE_VERSION=""
 ARG PULP_DEB_VERSION=""
 ARG PULP_RPM_VERSION=""
 ARG PULP_MAVEN_VERSION=""
+ARG PULP_PYTHON_VERSION="==3.0.0b12"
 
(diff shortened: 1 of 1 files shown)


bmbouter (commented):
This is great!

pulp_fedora31/Containerfile
```diff
@@ -7,6 +7,7 @@ ARG PULP_ANSIBLE_VERSION=""
 ARG PULP_DEB_VERSION=""
 ARG PULP_RPM_VERSION=""
 ARG PULP_MAVEN_VERSION=""
+ARG PULP_PYTHON_VERSION="==3.0.0b12"
```
daviddavis, line 10: Why are you specifically hardcoding the pulp_python version?

mikedep333 (approved):
:snake:

pulp_fedora31/Containerfile
mdellweg: I guess, because it has not been officially released yet.
And i also thought, we only included released Plugins here, because the upgrade path would be horrifying.

PR #36, opened 2020-12-11
Add pulp-certguard to fedora31 container

mdellweg:
[noissue]

This will fail, until pulp-certguard is released compatible with 3.9.

### DIFF
--- a/pulp_fedora31/Containerfile
+++ b/pulp_fedora31/Containerfile
@@ -1,22 +1,24 @@
 FROM pulp/pulp-ci:latest
 
 ARG PULPCORE_VERSION=""
-ARG PULP_FILE_VERSION=""
-ARG PULP_CONTAINER_VERSION=""
 ARG PULP_ANSIBLE_VERSION=""
+ARG PULP_CERTGUARD_VERSION=""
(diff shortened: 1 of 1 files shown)


mdellweg:
Oh, i was wrong. It forced pulpcore to go back to 3.8.1.
I cannot even decide which one is worse.

mdellweg:
Now with pulp-certguard released we should have it all set up.

fao89 (approved):

pulp_fedora31/Containerfile
```diff
@@ -16,7 +17,8 @@ RUN pip3 install --upgrade --use-feature=2020-resolver \
   pulp-ansible${PULP_ANSIBLE_VERSION} \
   pulp-deb${PULP_DEB_VERSION} \
   pulp-rpm${PULP_RPM_VERSION} \
-  pulp-maven${PULP_MAVEN_VERSION}
+  pulp-maven${PULP_MAVEN_VERSION} \
+  pulp-certguard${PULP_CERTGUARD_VERSION}
```
ekohl, line 21: You should sort the packages and put `requests` last. Then you can add `\` on the end of all plugins and have smaller diffs in the future.

PR #37, opened 2021-01-04
Update workflow to publish release images

daviddavis:
[noissue]

### DIFF
--- a/.github/workflows/publish_images.yaml
+++ b/.github/workflows/publish_images.yaml
@@ -7,18 +7,33 @@ on:
   schedule:
     # * is a special character in YAML so you have to quote this string
     - cron: '0 1 * * *'
+  workflow_dispatch:
+    inputs:
+      pulpcore:
+        description: 'Pulpcore Y-release Version (e.g. 3.7)'
+        required: true
 
 jobs:
   build:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Build the images
+      - name: Download s6-overlay
         run: |
           wget https://github.com/just-containers/s6-overlay/releases/download/v1.22.1.0/s6-overlay-amd64.tar.gz
-          docker build --file pulp_ci/Containerfile --tag pulp/pulp-ci:latest .
-          docker build --file pulp_fedora31/Containerfile --tag pulp/pulp-fedora31:latest .
-          docker build --file pulp_galaxy_ng/Containerfile --tag pulp/pulp-galaxy-ng:latest .
+      - name: Build the images
+        run: |
+          version="${{ github.event.inputs.pulpcore }}"
+          if [ -z "$version" ]; then
+            version="latest"
+            docker build --file pulp_ci/Containerfile --tag pulp/pulp-ci:latest .
+            docker build --file pulp_fedora31/Containerfile --tag pulp/pulp-fedora31:latest .
+            docker build --file pulp_galaxy_ng/Containerfile --tag pulp/pulp-galaxy-ng:latest .
+          else
+            version_req="$version.0"
+            docker build --file pulp_fedora31/Containerfile --tag pulp/pulp-fedora31:$version --build-arg PULPCORE_VERSION="~=$version_req" .
+            docker build --file pulp_galaxy_ng/Containerfile --tag pulp/pulp-galaxy-ng:$version --build-arg PULPCORE_VERSION="~=$version_req" .
+          fi
       - name: Docker login
         env:
           DOCKER_BOT_PASSWORD: ${{ secrets.DOCKER_BOT_PASSWORD }}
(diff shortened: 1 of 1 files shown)


.github/workflows/publish_release_images.yaml
```diff
@@ -0,0 +1,34 @@
+name: Publish Release OCI Images
+
+on:
+  push:
+    branches:
+      - master
```
daviddavis, line 6: I plan to remove this after we get some release images published to docker hub.

.github/workflows/publish_release_images.yaml
```diff
@@ -0,0 +1,34 @@
...
+  schedule:
+    # * is a special character in YAML so you have to quote this string
+    - cron: '0 3 * * 6'
```
daviddavis, line 9: It would be nice to be more intelligent about when to publish release images but for now, I think weekly is sufficient.

.github/workflows/publish_release_images.yaml
```diff
@@ -0,0 +1,34 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        pulpcore:
+          - "3.5"
+          - "3.6"
+          - "3.7"
+          - "3.8"
```
fao89, line 20: cool!

.github/workflows/publish_release_images.yaml
```diff
@@ -0,0 +1,34 @@
...
+          wget https://github.com/just-containers/s6-overlay/releases/download/v1.22.1.0/s6-overlay-amd64.tar.gz
+          docker build --file pulp_fedora31/Containerfile --tag pulp/pulp-fedora31:${{ matrix.pulpcore }} --build-arg PULPCORE_VERSION="==${{ matrix.pulpcore }}" .
+      - name: Docker login
+        env:
+          DOCKER_BOT_PASSWORD: ${{ secrets.DOCKER_BOT_PASSWORD }}
+          DOCKER_BOT_USERNAME: ${{ secrets.DOCKER_BOT_USERNAME }}
+        run: echo "$DOCKER_BOT_PASSWORD" | docker login -u "$DOCKER_BOT_USERNAME" --password-stdin docker.io
+      - name: Push ci image to dockerhub
+        run: |
+          docker push docker.io/pulp/pulp-fedora31:${{ matrix.pulpcore }}
```
fao89, line 34: shouldn't we do the same for `pulp-galaxy-ng`?

.github/workflows/publish_release_images.yaml
mdellweg: Is there a manually?

mdellweg: Is there a manually?

fao89: Manual option: https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

daviddavis: Will add a manual trigger. Should I keep the cron job or ditch it?

fao89: I think we can ditch it

.github/workflows/publish_release_images.yaml
mdellweg: Should we have a branch for each release?

.github/workflows/publish_release_images.yaml
mdellweg: absolutely!

.github/workflows/publish_release_images.yaml
mdellweg: We could trigger a webhook, whenever a plugin for a specific pulpcore version released. (Future PR idea...)

daviddavis:
I switched the workflow to use a manual trigger. :+1:

.github/workflows/publish_images.yaml
```diff
@@ -7,25 +7,43 @@ on:
...
+        run: |
+          version="${{ github.event.inputs.pulpcore }}"
+          if [ -z "$version" ]; then
+            version="latest"
+            docker build --file pulp_ci/Containerfile --tag pulp/pulp-ci:latest .
+            docker build --file pulp_fedora31/Containerfile --tag pulp/pulp-fedora31:latest .
+            docker build --file pulp_galaxy_ng/Containerfile --tag pulp/pulp-galaxy-ng:latest .
+          else
+            docker build --file pulp_fedora31/Containerfile --tag pulp/pulp-fedora31:$version --build-arg PULPCORE_VERSION="==$version" .
+            docker build --file pulp_galaxy_ng/Containerfile --tag pulp/pulp-galaxy-ng:$version --build-arg PULPCORE_VERSION="==$version" .
```
mdellweg, line 34: Can we make this in a way that pulp-fedora31:3.7 will install pulpcore 3.7.3?
Does pip understand `pulpcore~=3.7`?

daviddavis: It does although we'd want `pulpcore~=3.7.0` for pulpcore-fedora1:3.7. I updated the workflow.

.github/workflows/publish_release_images.yaml
daviddavis: Just saw this question: what would be the benefit of a branch for each release?

mdellweg (approved):

PR #38, opened 2021-01-04
Retry the status api when testing pulp container

daviddavis:
Attempting to fix a bug where the curl command sometimes fail.

[noissue]


mdellweg (approved):
another implementation of wait_for_pulp ;)

PR #39, opened 2021-01-06
Skip building galaxy-ng when pulpcore release is specified

daviddavis:
The pulp-galaxy-ng Containerfile uses specific version dependencies and can't be built against just any pulpcore release.

Eventually, I'm hoping to have a separate input for the galaxy_ng image but for now, just skip building it.

[noissue]


daviddavis:
Here's the failure when you try to build pulp-fedora31:3.6:

```
ERROR: Cannot install galaxy-ng==4.2.0rc2 and pulpcore~=3.6.0 because these package versions have conflicting dependencies.
```

dkliban (approved):

PR #40, opened 2021-01-13
Adds pulp-python

fao89:
[noissue]

### DIFF
--- a/pulp_fedora31/Containerfile
+++ b/pulp_fedora31/Containerfile
@@ -7,6 +7,7 @@ ARG PULP_CONTAINER_VERSION=""
 ARG PULP_DEB_VERSION=""
 ARG PULP_FILE_VERSION=""
 ARG PULP_MAVEN_VERSION=""
+ARG PULP_PYTHON_VERSION=""
 ARG PULP_RPM_VERSION=""
 
 RUN pip3 install --upgrade --use-feature=2020-resolver \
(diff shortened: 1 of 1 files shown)


mdellweg (commented):
Is this limiting our ability to build images for older versions of pulp?

@daviddavis do you know?

daviddavis:
@mdellweg indeed, that will be [an issue](https://github.com/pulp/pulp_python/blob/34420ba1aa486290b731cedf927a5bf54e5c8965/requirements.txt#L1). Is there a way to exclude pulp_python if pulpcore is < 3.9?

daviddavis:
@mdellweg I think at one point you mentioned branching pulp-oci-images? That might be an option.

daviddavis:
Here's what I am thinking: 
* Create branches 3.7, 3.8, 3.9, and latest (and ditch master). 
* Pull this change to add pulp_python into latest and 3.9.
* The nightly job will run against latest and publish latest. 
* The manual job will take a branch name and publish it using the branch name as the tag

WDYT? cc @dkliban

dkliban:
That sounds good to me @daviddavis

daviddavis:
Ok, here is the change to the workflow: https://github.com/pulp/pulp-oci-images/pull/41

Once we merge that and branch, I think we can merge this change to add pulp_python.

daviddavis (approved):

mdellweg (approved):

PR #41, opened 2021-01-13
Update the publish job to check out branch

daviddavis:
[noissue]

### DIFF
--- a/.github/workflows/publish_images.yaml
+++ b/.github/workflows/publish_images.yaml
@@ -3,14 +3,14 @@ name: Build and publish OCI Images
 on:
   push:
     branches:
-      - master
+      - latest
   schedule:
     # * is a special character in YAML so you have to quote this string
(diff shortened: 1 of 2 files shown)


.github/workflows/publish_images.yaml
```diff
@@ -3,35 +3,36 @@ name: Build and publish OCI Images
...
+        description: 'Release branch (e.g. 3.7)'
         required: true
 
 jobs:
   build:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.inputs.pulpcore }}
```
daviddavis, line 22: I'm hoping that if this is not set, it'll default to the default branch (latest).

fao89: this is the only thing I'm really unsure of, but let's test it"

fao89 (approved):

PR #44, opened 2021-01-13
Silence 2020-resolver warnings

daviddavis:
```
WARNING: --use-feature=2020-resolver no longer has any effect, since it is now the default dependency resolver in pip. This will become an error in pip 21.0.
```

[noissue]


mdellweg (approved):
About time. ;) 2020 is done. :ballot_box_with_check:

PR #50, opened 2021-01-14
Upgrade galaxy_ng to 4.2.1

bmbouter:

PR #51, opened 2021-01-15
Initial data for Galaxy

fao89:
This adds initial data for Galaxy:
```
[cont-init.d] pulp-galaxy-api: executing... 
Installed 21 object(s) from 1 fixture(s)
```
I'm wondering if we should trigger this as default, or just add the fixture to a path and the user runs:
```
/usr/local/bin/pulpcore-manager loaddata /etc/pulp-galaxy/initial_data.json
```

### DIFF
--- a/pulp_galaxy_ng/README.md
+++ b/pulp_galaxy_ng/README.md
@@ -1,0 +1,8 @@
+# Galaxy
+
+For loading the initial data:
+```bash
+$ DATA_FIXTURE_URL="https://raw.githubusercontent.com/ansible/galaxy_ng/master/dev/automation-hub/initial_data.json"
+$ curl $DATA_FIXTURE_URL | <docker | podman> exec -i pulp bash -c "cat > /tmp/initial_data.json"
+$ <docker | podman> exec pulp bash -c "/usr/local/bin/pulpcore-manager loaddata /tmp/initial_data.json"
+```


bmbouter:
I'm in favor of this, but it does create a maintenance concern where they change the assumption of the base data and this won't receive the update. I'm still in favor of merging it tho.

fao89:
> I'm in favor of this, but it does create a maintenance concern where they change the assumption of the base data and this won't receive the update. I'm still in favor of merging it tho.

I was thinking about this, and a possible solution would be a bash script:
```
/some/default/path/galaxy_data_script.sh
```
This script would have:
```bash
curl https://raw.githubusercontent.com/ansible/galaxy_ng/master/dev/automation-hub/initial_data.json > initial_data.json
/usr/local/bin/pulpcore-manager loaddata initial_data.json
```
And the user would run it when convenient
```
podman exec -it pulp /some/default/path/galaxy_data_script.sh   
```

bmbouter:
this is a great idea, can we do that in this PR instead!?

fao89:
> this is a great idea, can we do that in this PR instead!?

Yes, I'll do it

bmbouter:
@fao89 I think we'll keep 0.5.5 yes? https://github.com/pulp/pulp-oci-images/pull/50#discussion_r558380364

dkliban (approved):

pulp_galaxy_ng/Containerfile
```diff
@@ -15,3 +15,6 @@ RUN pip3 install --upgrade \
 RUN ln /usr/local/lib/python3.7/site-packages/pulp_ansible/app/webserver_snippets/nginx.conf /etc/nginx/pulp/pulp_ansible.conf
 RUN ln /usr/local/lib/python3.7/site-packages/pulp_container/app/webserver_snippets/nginx.conf /etc/nginx/pulp/pulp_container.conf
 RUN ln /usr/local/lib/python3.7/site-packages/galaxy_ng/app/webserver_snippets/nginx.conf /etc/nginx/pulp/galaxy_ng.conf
+
+RUN mkdir -p /etc/pulp-galaxy
+COPY pulp_galaxy_ng/assets/initial_data.sh /etc/pulp-galaxy/initial_data.sh
```
fao89: I realize it is more a documentation issue:

```
~ via 🐍 v3.8.0 (venv) 
❯ curl https://raw.githubusercontent.com/ansible/galaxy_ng/master/dev/automation-hub/initial_data.json | podman exec -i pulp bash -c "cat > /tmp/initial_data.json"
podman exec pulp bash -c "/usr/local/bin/pulpcore-manager loaddata /tmp/initial_data.json"
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  6168  100  6168    0     0  93454      0 --:--:-- --:--:-- --:--:-- 93454
Installed 21 object(s) from 1 fixture(s)

```

PR #62, opened 2021-02-05
Set pulpcore version to ~3.9.0 for the 3.9 branch

dkliban:
[noissue]


goosemania (approved):

mdellweg (approved):
The fedora box already has it...

PR #73, opened 2021-04-23
Run pulp container on https

fao89:
Related thread: https://listman.redhat.com/archives/pulp-dev/2021-April/msg00043.html

### DIFF
--- a/.github/workflows/publish_images.yaml
+++ b/.github/workflows/publish_images.yaml
@@ -27,6 +27,13 @@ jobs:
           if [ "$VERSION" = "latest" ]; then
             docker build --file pulp_ci_centos/Containerfile --tag pulp/pulp-ci-centos:latest .
             docker build --file pulp_galaxy_ng/Containerfile --tag pulp/pulp-galaxy-ng:latest .
+            # SSL
+            docker build --file pulp_ci_centos/SSL/Containerfile --tag pulp/pulp-ci-centos:https .
+            sed -i "s/latest/https/g" pulp_galaxy_ng/Containerfile
+            docker build --file pulp_galaxy_ng/Containerfile --tag pulp/pulp-galaxy-ng:https .
+            sed -i "s/latest/https/g" pulp/Containerfile
+            docker build --file pulp/Containerfile --tag pulp/pulp:https .
+            sed -i "s/https/latest/g" pulp/Containerfile
             # Be resilient if this fails
             docker build --file pulp_nightly/Containerfile --tag pulp/pulp:nightly . || true
           fi
@@ -43,6 +50,9 @@ jobs:
           if [ "$VERSION" = "latest" ]; then
             docker push docker.io/pulp/pulp-ci-centos:latest
             docker push docker.io/pulp/pulp-galaxy-ng:latest
+            docker push docker.io/pulp/pulp-ci-centos:https
+            docker push docker.io/pulp/pulp-galaxy-ng:https
+            docker push docker.io/pulp/pulp:https
             # Be resilient if this fails
             docker push docker.io/pulp/pulp:nightly || true
           fi
--- a/pulp_ci_centos/SSL/Containerfile
+++ b/pulp_ci_centos/SSL/Containerfile
@@ -1,0 +1,8 @@
+FROM pulp/pulp-ci-centos:latest
+
+RUN dnf -y install python3-cryptography
+
+COPY assets/SSL/openssl.cnf /etc/ssl/pulp/openssl.cnf
+COPY assets/SSL/v3.cnf /etc/ssl/pulp/v3.cnf
+COPY assets/SSL/pulpcore-api.prep /etc/cont-init.d/pulpcore-api
+COPY assets/SSL/nginx.conf /etc/nginx/nginx.conf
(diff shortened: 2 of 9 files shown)


assets/pulpcore-api.prep
```diff
@@ -8,5 +8,13 @@ foreground {
...
+  openssl genrsa -out /etc/pulp/certs/ca.key 2048
+  openssl req -x509 -new -nodes \
+    -key /etc/pulp/certs/ca.key -subj "/CN=pulp/C=US/L=California" \
+    -out /etc/pulp/certs/ca.crt
+  openssl genrsa -out /etc/pulp/certs/pulp_webserver.key 2048
+  openssl req -new -key /etc/pulp/certs/pulp_webserver.key -subj "/CN=pulp/C=US/ST=California" -out /etc/pulp/certs/pulp_webserver.csr
 }
 openssl ec -in /etc/pulp/certs/token_private_key.pem -pubout -out /etc/pulp/certs/token_public_key.pem
+openssl x509 -req -in /etc/pulp/certs/pulp_webserver.csr -CA /etc/pulp/certs/ca.crt -CAkey /etc/pulp/certs/ca.key \
+    -CAcreateserial -out /etc/pulp/certs/pulp_webserver.crt
```
mdellweg: That is not how execlineb works...
You need to `foreground` every single command that does not support chain calling, despite the last one.

fao89: I got confused when reading about `foreground`, thank you for the help!

mdellweg:
I believe, you need to set verify-ssl or similar to false for the curl calls.

daviddavis:
This work looks great. Can I ask for a huge favor? Can we wait until after April 30 to merge this work (or at least hold off publishing the new containers until then)? The students working on the UI are using these and have their final project deadline on April 30. I don't want to mess them up.

fao89:
> This work looks great. Can I ask for a huge favor? Can we wait until after April 30 to merge this work (or at least hold off publishing the new containers until then)? The students working on the UI are using these and have their final project deadline on April 30. I don't want to mess them up.

yes, this is just a POC, and it may break CI for almost every plugin. It needs to be proper advertised before merged, so people can have time to improve functional tests to deal with https

fao89:
Related thread: https://listman.redhat.com/archives/pulp-dev/2021-April/msg00043.html

mdellweg:
This is a major change in the way the single container works.
Can we somehow make the switch to https optional? So we do not break every automation we have when merging this?

daviddavis:
We created [an https tag](https://hub.docker.com/layers/pulp/pulp/https/images/sha256-8467255764ede6748eb1991614345dc42c6b740d57dde9d19457f166179b77ea?context=explore) that @fao89 is using to test things out. 

Ideally I think we'd like to get all the other tags moved over to https but this seems like a monumental undertaking when one considers all the repos, release branches, etc.

mdellweg:
What about the opposite tag (http / nossl / ...), and some day, we announce the switchover of latest?
Can we make the ssl version to be a thin container layer on top of the nossl version?

fao89:
Followed: https://docs.oracle.com/cd/E52668_01/E66514/html/ceph-issues-24424028.html

pulp_ci_centos/SSL/Containerfile
```diff
@@ -0,0 +1,8 @@
+FROM pulp/pulp-ci-centos:latest
```
mdellweg, line 1: :100:

.github/workflows/publish_images.yaml
```diff
@@ -37,6 +37,13 @@ jobs:
           if [ "$VERSION" = "latest" ]; then
             docker build --file pulp_ci_centos/Containerfile --tag pulp/pulp-ci-centos:latest .
             docker build --file pulp_galaxy_ng/Containerfile --tag pulp/pulp-galaxy-ng:latest .
+            # SSL
+            docker build --file pulp_ci_centos/SSL/Containerfile --tag pulp/pulp-ci-centos:https .
+            sed -i "s/latest/https/g" pulp_galaxy_ng/Containerfile
```
mdellweg, line 32: This is changing the files checked into this repository.
May be OK in CI, but the moment we are going to write up these steps in the README, it may get weird. WDYT?

himdel:
I'd like to request this feature be optional, with an option to go back to http.

We're using this for Cypress testing, and having to generate a self-signed certificate only to ignore it seems unecessary for that case, if it can be reasonably avoided :).

(context: https://github.com/ansible/ansible-hub-ui/pull/378#discussion_r629397315 )

.github/workflows/publish_images.yaml
fao89: I didn't want to have a new container file just because of the tag, maybe we need to use a template

fao89: template + makefile?

mdellweg: So you would turn the SSL/Containerfile into a template? Would you template it on the fly? Or use the Makefile to template files we want to check into git? I really don't know what's the best way.
Another thought: We are now building the images with 3 layers of dependency. Maybe it _is_ time to start using make to sort out the build sequence.
BTW: we could call that file `pulp-ci-centos8/Containerfile.https`. No one really says that it must be called `Containerfile`. So one nice naming schema would be `<image_name>/Containerfile.<tag>`. Wdyt?

fao89: I agree with everything you said, including the part: "I really don't know what's the best way"
`<image_name>/Containerfile.<tag>`sounds really good!

PR #75, opened 2021-05-03
Remove issue checking

daviddavis:
We plan to no longer tracking issues on redmine and we don't generate a changelog, etc so I think there's no much value in checking commits. I propose we remove issue checking from this repo.


mdellweg (approved):
I agree. Most of the changes are on the version branches anyway.

PR #83, opened 2021-05-17
Use new worker entrypoint

mdellweg:
To still support versionsi prior to 3.13, keep the old startup scripts around.

[noissue]

PR #89, opened 2021-06-07
Revert "Use new worker entrypoint"

dkliban:
This reverts commit 53ff47ccb0b13d9b1c69377466ef8443b4670b8f.

[noissue]


daviddavis (approved):

mdellweg:
Do we know what the issue with this commit is?
We need a change like this in any case.

daviddavis:
There’s some more info here: https://github.com/pulp/pulp-oci-images/pull/83#issuecomment-856157593

PR #91, opened 2021-06-08
Create the db encryption key

daviddavis:
fixes #8731

### DIFF
--- a/assets/pulpcore-api.prep
+++ b/assets/pulpcore-api.prep
@@ -1,10 +1,19 @@
 #!/usr/bin/execlineb -P
 
-if -n {
-  ls /etc/pulp/certs
+foreground {
+  if -n {
+    ls /etc/pulp/certs
+  }
+  mkdir -p /etc/pulp/certs
 }
 foreground {
-  mkdir -p /etc/pulp/certs
+  if -n {
+    ls /etc/pulp/certs/database_fields.symmetric.key
+  }
+  openssl rand -base64 32 > /etc/pulp/certs/database_fields.symmetric.key
+}
+if -n {
+  ls /etc/pulp/certs/token_private_key.pem
 }
 foreground {
   openssl ecparam -genkey -name prime256v1 -noout -out /etc/pulp/certs/token_private_key.pem


assets/pulpcore-api.prep
```diff
@@ -6,6 +6,9 @@ if -n {
 foreground {
   mkdir -p /etc/pulp/certs
 }
+foreground {
+  dd if=/dev/urandom bs=32 count=1 2>/dev/null | openssl base64 > /etc/pulp/certs/database_fields.symmetric.key
```
mdellweg: not sure, urandom is a good choice for an encryption key.
Is `openssl base64` any different from `base64`?
The rest looks good to me.

daviddavis: Here's another option I discovered?

```
openssl rand -base64 32
```

mikedep333: I was researching how to generate it as securely as possible.

Based on my research, that sounds like a more secure manner:
[OpenSSL 1.1.1d behavior](https://manpages.debian.org/buster/openssl/openssl-rand.1ssl.en.html)
[OpenSSL 3.0 Alpha behavior](https://manpages.debian.org/experimental/openssl/openssl-rand.1ssl.en.html)

Might I suggest we ask the Fenret project for advice, since this is security-related, and they are the experts.

daviddavis: The only fernet presence I see online is at [this repo for their spec](https://github.com/fernet/spec/issues) but it looks pretty dead. I have a couple contacts on the security team at Red Hat. Let me reach out to them for advice.

assets/pulpcore-api.prep
```diff
@@ -6,6 +6,9 @@ if -n {
 foreground {
   mkdir -p /etc/pulp/certs
 }
```
mikedep333: Users may want to provide the db encryption key but not the token key.

So please use a separate checks for each of these, and create if they do not exist:
1. /etc/pulp/certs dir
2. the database key
3. the token key.

assets/pulpcore-api.prep
mikedep333: Please use the same filepath as the installer:
`/etc/pulp/certs/database_fields.symmetric.key`
https://pulp.plan.io/issues/8192#note-18

mikedep333 (changes requested):

assets/pulpcore-api.prep
daviddavis: Isn't this the same filepath?

daviddavis:
Ok, updated.

assets/pulpcore-api.prep
```diff
@@ -6,6 +6,15 @@ if -n {
...
+}
+foreground {
+  openssl rand -base64 32 > /etc/pulp/certs/database_fields.symmetric.key
+}
+if -n {
+  ls /etc/pulp/certs/token_private_key.pem
+}
 foreground {
   openssl ecparam -genkey -name prime256v1 -noout -out /etc/pulp/certs/token_private_key.pem
 }
```
mdellweg: I think, the foreground needs to wrap your if construct. See example here:

https://skarnet.org/software/execline/if.html

daviddavis: Does [this if](https://github.com/pulp/pulp-oci-images/blob/latest/assets/pulpcore-api.prep#L3-L5) need to be wrapped in a foreground construct too?

mdellweg: Let's say, the way it's written now,i think it runs all three subsequent commands like they were in a block.

execline is weird!

daviddavis: Yea, I read through the docs and I am still not sure I understand it completely. I pushed a change that I hope is correct?

assets/pulpcore-api.prep
```diff
@@ -1,12 +1,21 @@
...
+  }
+  openssl rand -base64 32 > /etc/pulp/certs/database_fields.symmetric.key
+}
+foreground {
+  if -n {
+    ls /etc/pulp/certs/token_private_key.pem
+  }
   openssl ecparam -genkey -name prime256v1 -noout -out /etc/pulp/certs/token_private_key.pem
 }
 openssl ec -in /etc/pulp/certs/token_private_key.pem -pubout -out /etc/pulp/certs/token_public_key.pem
```
mdellweg: I think we should still consider the last two commands as a "block".

```suggestion
if -n {
  ls /etc/pulp/certs/token_private_key.pem
}
foreground {
  openssl ecparam -genkey -name prime256v1 -noout -out /etc/pulp/certs/token_private_key.pem
}
openssl ec -in /etc/pulp/certs/token_private_key.pem -pubout -out /etc/pulp/certs/token_public_key.pem
```

mdellweg (approved):
I think that is it.

assets/pulpcore-api.prep
```diff
@@ -1,11 +1,20 @@
...
+    ls /etc/pulp/certs/database_fields.symmetric.key
+  }
+  openssl rand -base64 32 > /etc/pulp/certs/database_fields.symmetric.key
+}
+if -n {
+  ls /etc/pulp/certs/token_private_key.pem
+}
 foreground {
   openssl ecparam -genkey -name prime256v1 -noout -out /etc/pulp/certs/token_private_key.pem
 }
```
mikedep333, line 20: The line to generate the public key should also be part of this block.

It looks like it's only a public key and not a certificate, but it's still not a good practice to regenerate an identical key over and over.

mikedep333 (changes requested):

assets/pulpcore-api.prep
mdellweg: It is part of the "block" due to the weird syntax of execline.
The `if` in line 15 is either exiting the script immediately, or tail call into the rest.

Don't confuse the `{}` with block structure here.

mikedep333: Wow. thank you!

mikedep333 (approved):

PR #94, opened 2021-06-24
Use new worker entrypoint

mdellweg:
To still support versions prior to 3.13, keep the old startup scripts around.


mdellweg:
This is a simplified version of https://github.com/pulp/pulp-oci-images/pull/83
It no longer tries to be clever and shut down redis or the resource-manager.

dralley (approved):

PR #95, opened 2021-06-25
Fix missing new-pulpcore-resource-manager/run

mdellweg:
The service file for the resource-manager with the new entry point was
missing.

[noissue]


daviddavis (approved):

dralley (approved):

dralley:
Does the old one not need to be removed?

pulp_ci_centos/Containerfile
```diff
@@ -83,6 +83,7 @@ COPY assets/postgres.run /etc/services.d/postgresql/run
 COPY assets/redis.run /etc/services.d/redis/run
 COPY assets/pulpcore-worker.prep /etc/cont-init.d/pulpcore-worker
 COPY assets/pulpcore-resource-manager.run /etc/services.d/pulpcore-resource-manager/run
```
dralley, line 85: Does the old one not need to be removed?

mdellweg: No, they are deactivated base on the pulpcore version.
We still want to be able to release old branches, right?

PR #103, opened 2021-07-14
Disable python36 and install python38

daviddavis:
[noissue]

PR #104, opened 2021-07-14
Install packaging package

daviddavis:
It's required by the pulp worker prep file:

https://github.com/pulp/pulp-oci-images/blob/33264a925d2c7e0afcad4ed9283a21978832ae89/assets/pulpcore-worker.prep#L16

[noissue]


daviddavis:
It got removed when [I stopped installing all of the createrepo_c deps](https://github.com/pulp/pulp-oci-images/pull/103) (as doing so raised an error).

fao89 (approved):

PR #105, opened 2021-07-14
Run pulp container on https

fao89:
[noissue]

### DIFF
--- a/.github/workflows/pr_build_images.yaml
+++ b/.github/workflows/pr_build_images.yaml
@@ -30,3 +30,30 @@ jobs:
       - name: Display log on error
         if: failure()
         run: docker logs pulp
+
+  sslbuild:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - ci_type: pulp_ci_centos
+            ci_image: pulp-ci-centos
+            test_type: pulp_galaxy_ng
+            test_image: pulp-galaxy-ng
+          - ci_type: pulp_ci_centos
+            ci_image: pulp-ci-centos
+            test_type: pulp
+            test_image: pulp
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build the images
+        run: |
+          wget https://github.com/just-containers/s6-overlay/releases/download/v1.22.1.0/s6-overlay-amd64.tar.gz
+          docker build --build-arg SCHEME="https" --file ${{ matrix.ci_type }}/Containerfile --tag pulp/${{ matrix.ci_image }}:https .
+          docker build --build-arg FROM_TAG="https" --file ${{ matrix.test_type }}/Containerfile --tag pulp/${{ matrix.test_image }}:https .
+      - name: Test the images
+        run: assets/test.sh "pulp/${{ matrix.test_image }}:https" https
+      - name: Display log on error
+        if: failure()
+        run: docker logs pulp
--- a/.github/workflows/publish_images.yaml
+++ b/.github/workflows/publish_images.yaml
@@ -50,3 +50,49 @@ jobs:
       - name: Display log on error
         if: failure()
         run: docker logs pulp
+
+  ssl_publish:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Download s6-overlay
+        run: |
+          wget https://github.com/just-containers/s6-overlay/releases/download/v1.22.1.0/s6-overlay-amd64.tar.gz
+      - name: Set version
+        run: |
+          version="${GITHUB_REF#refs/heads/}-https"
+          echo "Building and publishing $version."
+          echo "VERSION=$version" >> $GITHUB_ENV
+      - name: Build the images
+        run: |
+          docker build --build-arg SCHEME="https" --file pulp_ci_centos/Containerfile --tag pulp/pulp-ci-centos:https .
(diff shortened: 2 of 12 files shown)


fao89:
it seems latest branch is not protected

daviddavis:
The changes to the CI files here look good to me.

I'm hoping someone else can review the other files like the nginx conf stuff since it's beyond my level of sysadmin expertise.

fao89:
> The changes to the CI files here look good to me.
> 
> I'm hoping someone else can review the other files like the nginx conf stuff since it's beyond my level of sysadmin expertise.

the nginx conf comes from pulp_installer

.github/workflows/pr_build_images.yaml
```diff
@@ -30,3 +30,30 @@ jobs:
...
+          - ci_type: pulp_ci_centos
+            ci_image: pulp-ci-centos
+            test_type: pulp
+            test_image: pulp
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build the images
+        run: |
+          wget https://github.com/just-containers/s6-overlay/releases/download/v1.22.1.0/s6-overlay-amd64.tar.gz
+          docker build --build-arg SCHEME="https" --file ${{ matrix.ci_type }}/Containerfile --tag pulp/${{ matrix.ci_image }}:latest .
```
mdellweg: This is subtle.
Does the 'pulp' image use the latest tag for it's base image?

.github/workflows/publish_images.yaml
```diff
@@ -50,3 +50,50 @@ jobs:
...
+      - name: Build the images
+        run: |
+          docker build --build-arg SCHEME="https" --file pulp_ci_centos/Containerfile --tag pulp/pulp-ci-centos:latest .
+          if [ "$VERSION" = "latest-https" ]; then
+            docker build --build-arg SCHEME="https" --file pulp_ci_centos/Containerfile --tag pulp/pulp-ci-centos:https .
+            docker build --file pulp_galaxy_ng/Containerfile --tag pulp/pulp-galaxy-ng:https .
+            docker build --file pulp_galaxy_ng/Containerfile --tag pulp/pulp:https .
+          else
+            docker build --file pulp/Containerfile --tag pulp/pulp:$VERSION .
+          fi
```
mdellweg: This looks wrong somehow. Is it right, that we need the ssl image to be called "latest" for this build?
But why build the same thing again in line 70 instead of just "docker tag" it to the new name.

fao89: yes, it is hacky because the images point to latest
```
FROM pulp/pulp-ci-centos:latest
```
and I agree, I change the PR to use docker tag

.github/workflows/pr_build_images.yaml
mdellweg: Maybe a dumb question, but can we use the `build-arg` to specify the tag of the base image in a Container file?
I have an uneasy feeling with calling the images with the wrong tag to serve the build process.

pulp_ci_centos/Containerfile
```diff
@@ -22,6 +22,8 @@ ENV BUILDAH_ISOLATION=chroot
 RUN dnf -y module disable python36
 RUN dnf -y module enable python38
 
+ARG SCHEME="http"
+
```
mdellweg: Is it possible to move this to the very top of the file?

pulp_nightly/Containerfile
```diff
@@ -1,4 +1,5 @@
-FROM pulp/pulp-ci-centos:latest
+ARG FROM_TAG="latest"
+FROM pulp/pulp-ci-centos:${FROM_TAG}
```
mdellweg, line 2: :heart:

mdellweg: This looks like the cleanest approach to me so far.
Maybe this is also the way to forward with images for different python versions.

pulp_ci_centos/Containerfile
```diff
@@ -98,8 +104,12 @@ COPY assets/pulpcore-api.prep /etc/cont-init.d/pulpcore-api
 COPY assets/pulpcore-api.run /etc/services.d/pulpcore-api/run
 COPY assets/000-postgres.prep /etc/cont-init.d/000-postgres
 COPY assets/zzz-postgres.prep /etc/cont-init.d/zzz-postgres
-COPY assets/nginx.conf /etc/nginx/nginx.conf
+COPY assets/ssl_nginx.conf /nginx/ssl_nginx.conf
+COPY assets/nginx.conf /nginx/nginx.conf
 COPY assets/nginx.run /etc/services.d/nginx/run
+COPY assets/webserver.sh /nginx/webserver.sh
+
+RUN /nginx/webserver.sh
```
mdellweg, line 112: This will use the $SCHEME buildarg.
Maybe make it a positional argument.

Also would it be possible to not perform all the ssl setup (especially the cert generation) in the http container?

mdellweg (approved):
My review is mostly concerning the CI and Containerfiles. I have no idea what it takes to make nginx use ssl (i heard the files come from the installer).
That being said, this is great work. And i like that we have separate images for http and https.
Maybe consider my suggestions as such for a follow up PR.

pulp_ci_centos/Containerfile
fao89: I'll be merging this PR and I'll try to address that on a follow-up PR

PR #106, opened 2021-07-22
Increase max connections

fao89:
Ref: https://github.com/pulp/community/discussions/55
[noissue]

### DIFF
--- a/assets/000-postgres.prep
+++ b/assets/000-postgres.prep
@@ -26,6 +26,7 @@ backtick -n ! { pipeline { postgres --version } egrep -o "[0-9]{1,}\.[0-9]{1,}" 
 importas -u PGVERSION !
 define PGHOME "/var/lib/pgsql"
 define PGDATA "${PGHOME}/data"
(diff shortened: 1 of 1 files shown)


assets/000-postgres.prep
```diff
@@ -52,6 +53,9 @@ ifthenelse { s6-test -d "${PGDATA}/base" -a -f "${PGDATA}/PG_VERSION" }
   }
 }
 }
+foreground {
+  sed -i "s/max_connections = 100/max_connections = 1024/g" ${PG_CONF_PATH}
```
mdellweg, line 57: Any chance we can do this at image build time?

fao89: I tried https://github.com/pulp/pulp_ansible/pull/599
but when you try to restart postgres it crashes

mdellweg: OH, i meant somewhere in the Containerfile but maybe that file is only created on the fly.

fao89:
Closing in favor of https://github.com/pulp/pulp_ansible/pull/599

PR #108, opened 2021-07-26
Use python36 container for release builds

mdellweg:
[noissue]

I want to build a 3.11 container with the latest z-stream of pulp_rpm, and i think this is, what it takes.


fao89 (commented):
looks good, I don't get why CI is failing

mdellweg:
> looks good, I don't get why CI is failing

`ValueError: 'default' must be a list when 'multiple' is true.` That is click 7/8 compatibility...

mdellweg:
But i'm also unsure if the PR CI is testing the right stuff for a release branch...

PR #110, opened 2021-07-26
Bump pulp_rpm version on the 3.14 release branch

dralley:
[noissue]


pulp/Containerfile
```diff
@@ -8,7 +8,7 @@ ARG PULP_DEB_VERSION=""
 ARG PULP_FILE_VERSION=""
 ARG PULP_MAVEN_VERSION=""
 ARG PULP_PYTHON_VERSION=""
-ARG PULP_RPM_VERSION="~=3.13.2"
+ARG PULP_RPM_VERSION="~=3.14.0"
```
dralley, line 11: Should I get rid of this entirely?  Or should we fill out the required versions for all other plugins?

mdellweg: I have had the same question earlier that day. For consistent builds, pinning all plugins to y (as done here) would be good.
This probably depends on whether we want to use those containers in integration tests.

mdellweg (approved):

PR #111, opened 2021-07-26
Fix issue with arg scheme

fao89:
https://ryandaniels.ca/blog/docker-dockerfile-arg-from-arg-trouble/
[noissue]


pulp_ci_centos/Containerfile
```diff
@@ -1,5 +1,7 @@
-ARG SCHEME="http"
 FROM centos:8
+# https://ryandaniels.ca/blog/docker-dockerfile-arg-from-arg-trouble/
```
mdellweg, line 2: OMG

PR #112, opened 2021-07-26
Fix 3.14 build

fao89:
[noissue]


fao89:
@dralley build broke because it was getting `pulp/pulp-ci-centos:latest` which is python 3.8 and building `pulp:3.14` which is python 3.6
https://github.com/pulp/pulp-oci-images/actions/runs/1068846529

daviddavis (approved):

dralley (approved):

PR #115, opened 2021-08-03
Create db encryption key before migrating

### LINKED ISSUE #114: DB encryption key is created after migrating
It needs to be created before we migrate.

daviddavis:
fixes #114


dkliban (approved):

fao89 (approved):
I probably need to change my pulp-operator PR

PR #128, opened 2021-11-02
Pin pulp-file

fao89:
[noissue]


bmbouter (approved):
Looks good to me, thanks!

dkliban (approved):

pulp/Containerfile
```diff
@@ -6,7 +6,7 @@ ARG PULP_ANSIBLE_VERSION=""
 ARG PULP_CERTGUARD_VERSION=""
 ARG PULP_CONTAINER_VERSION=""
 ARG PULP_DEB_VERSION=""
-ARG PULP_FILE_VERSION=""
+ARG PULP_FILE_VERSION="<1.10"
```
mdellweg, line 9: Hmm. I don't think, this is the soluttion here. I think we should make 1.10 unavailable on pypi (it has broken deps).
And if we want to fix anything here, make it `PULP_FILE_VERSION="!1.10"`.

bmbouter:
After some matrix discussion, we decided to yank 1.10.0 (which I did) and @fao89 will revert this PR and rerun the OCI publish.

PR #130, opened 2021-11-04
Rebuild and publish containers on a regular basis

### LINKED ISSUE #123: Rebuild and publish containers on a regular basis
When a new plugin gets released, the single container does not get updated automatically. For now, we should rebuild supported containers automatically on a regular basis (e.g. weekly).

fao89:
closes #123


mdellweg:
Nice one. However, i kind of remember from the discussion, we wanted to curate a list of releases to rebuild regularly. Also maybe this should be a separate workflow, because "every day" could be too much.
@dkliban can you help my memory here?

bmbouter:
+1 to this as is. I like that it's nightly. I also like that it's super simple and doesn't try to be smart about what it's publishing it just goes.

dkliban (approved):

PR #134, opened 2021-12-03
Change client_max_body_size to 1024m

mnasiadka:
Current limit - 10m doesn't allow to upload bigger RPM artifacts.

### DIFF
--- a/assets/nginx.conf
+++ b/assets/nginx.conf
@@ -59,7 +59,7 @@ http {
             # redirects, we set the Host: header above already.
             proxy_redirect off;
             proxy_pass http://pulp-api;
-            client_max_body_size 1024m;
+            client_max_body_size 0;
         }
 
         location /auth/login/ {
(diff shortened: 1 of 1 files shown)


assets/nginx.conf
```diff
@@ -59,7 +59,7 @@ http {
             # redirects, we set the Host: header above already.
             proxy_redirect off;
             proxy_pass http://pulp-api;
-            client_max_body_size 10m;
+            client_max_body_size 1024m;
```
fao89: maybe we should disable it
```suggestion
            client_max_body_size 0;
```
https://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size

ipanova: +1 rpm can very much vary in size

fao89 (approved):

mdellweg:
The idea was that you can still use the chunked upload.

PR #135, opened 2021-12-08
Removes resource manager from single container

bmbouter:
The resource manager no longer able to be run in the upcoming version of
Pulp.

### DIFF
--- a/assets/new-pulpcore-resource-manager.run
+++ b/assets/new-pulpcore-resource-manager.run
@@ -1,5 +1,0 @@
-#!/usr/bin/execlineb -P
-export DJANGO_SETTINGS_MODULE pulpcore.app.settings
-export PULP_SETTINGS /etc/pulp/settings.py
-export PATH /usr/local/bin:/usr/bin/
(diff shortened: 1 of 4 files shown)


mdellweg:
Do we already use a versioned variant of the `pulp_ci_centos` image in the older release branches? If not, this will break everything still depending on the resource-manager running.

bmbouter:
I thought this over some, and I think we should merge this even if later we'll have to put in more work to release older branches. Mainly I just think we need to get this ready for the 3.17 release.

mdellweg:
> I thought this over some, and I think we should merge this even if later we'll have to put in more work to release older branches. Mainly I just think we need to get this ready for the 3.17 release.

We need to be able to produce z stream releases at any point in time. So i think we need a strategy.

For now, we can extend the version checking strategy and only start the resource-manager for pulpcore < 3.16.
I know ite in readline is a pain.

bmbouter:
I'm closing this PR in favor of https://github.com/pulp/pulp-oci-images/pull/136

PR #140, opened 2022-01-10
Restart pulpcore-api after migration

fao89:
[noissue]

### DIFF
--- a/assets/000-postgres.prep
+++ b/assets/000-postgres.prep
@@ -117,3 +117,15 @@ foreground {
   s6-setuidgid postgres
   /usr/local/bin/pulpcore-manager migrate
 }
+foreground {
+  s6-svc -r /var/run/s6/services/pulpcore-api
+}
+foreground {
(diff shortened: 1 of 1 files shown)


assets/000-postgres.prep
```diff
@@ -116,4 +116,5 @@ foreground {
   export PULP_CONTENT_ORIGIN localhost
   s6-setuidgid postgres
   /usr/local/bin/pulpcore-manager migrate
+  s6-svc -r /var/run/s6/services/pulpcore-api
```
mdellweg, line 121: Don't we start all the pulpcore services after the postgres service is up? At that time i expect the prep script to be executed already.
And if there _is_ a need to restart that service, why is there none for the others?

fao89: I'm testing why migrations fail on pulp_installer and not on a single container,
my theory is that we have the restart all services on the installer so it can see `permissions_assignment` doesn't exist anymore

https://github.com/pulp/pulp_container/runs/4742806544?check_suite_focus=true#step:13:366

PR #141, opened 2022-01-11
Ensure migrations run before pulp services

fao89:


assets/wait_on_database_migrations.sh
```diff
@@ -0,0 +1,21 @@
...
+    sleep 5
+  fi
+done
+
+if [ $database_migrated = false ]; then
+  echo "Database not migrated in time, exiting"
+  exit 1
+else
+  exit 0
+fi
```
fao89: got from pulp-operator: https://github.com/pulp/pulp-operator/blob/main/containers/images/pulp/container-assets/wait_on_database_migrations.sh

dkliban (approved):

assets/000-postgres.prep
```diff
@@ -117,3 +117,7 @@ foreground {
   s6-setuidgid postgres
   /usr/local/bin/pulpcore-manager migrate
 }
+pipeline {
+  echo "ready"
+}
+tee /database/status
```
mdellweg, line 123: This does not seem to be used. (I would have preferred that method.)

fao89: it is used here: https://github.com/pulp/pulp-oci-images/blob/latest/assets/wait_on_database_migrations.sh#L11

mdellweg: Oh right, but that line has no consequences for any of the services.

PR #144, opened 2022-01-25
Push to Github Container Registry

fao89:
https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry

[noissue]


ipanova:
cool stuff!

fao89:
I would like to experiment using github container registry on our CI to avoid using docker
@pulp/continuous-integration wdyt?

fao89:
tested here https://github.com/pulp/pulp_npm/pull/147

PR #157, opened 2022-01-31
Update CentOS image

fao89:


mdellweg (approved):
This means we pull from quay and push to ghr. I'm not saying this is a bad idea, just stating the obvious.

fao89:
the stream images are only published on quay
https://www.centos.org/centos-stream/

PR #162, opened 2022-02-10
All content on nightly

fao89:
[noissue]


mdellweg:
As a heavy user of that nightly container, i am weary of that change. But for the project as a whole, we should just do it.

mdellweg (approved):

dralley (approved):

PR #163, opened 2022-02-15
Renames new-pulpcore-worker@ services to pulpcore-worker@

dkliban:
[noissue]


fao89 (commented):
I think this PR is good, but I believe we should work on pulp-ci-centos tags first

mdellweg:
There are some prepare scripts that need deletion too. Will try to find them.

edit: Here they are:
- https://github.com/pulp/pulp-oci-images/blob/latest/assets/init/pulpcore-resource-manager
- https://github.com/pulp/pulp-oci-images/blob/latest/assets/init/pulpcore-worker

fao89:
> I think this PR is good, but I believe we should work on pulp-ci-centos tags first

https://github.com/pulp/pulp-oci-images/pull/164

fao89 (approved):

mdellweg (approved):

PR #188, opened 2022-05-12
Install skopeo

lubosmj:
[noissue]


fao89 (approved):

mdellweg (approved):

pulp_ci_centos/Containerfile
```diff
@@ -59,6 +59,7 @@ RUN dnf -y install dnf-plugins-core && \
     dnf -y install cairo-devel cmake gobject-introspection-devel cairo-gobject-devel && \
     dnf -y install libcurl-devel libxml2-devel sqlite-devel file-devel && \
     dnf -y install ostree-libs ostree && \
+    dnf -y install skopeo && \
```
fao89, line 62: do we need it on pulp-operator?
https://github.com/pulp/pulp-operator/blob/main/containers/images/pulp/Containerfile.core.j2

mdellweg: It is used in a suggested version of the container signing script as far as i know. So "need" is hard to pin down.

PR #201, opened 2022-06-07
Adding containerfile for pulp with debuginfod

RyanSGoldberg:
Added in containerfile which starts up [debuginfod](https://sourceware.org/elfutils/Debuginfod.html) server as background process. Makes it easy for debugger-like tools, such as valgrind and gdb, run both in and out of the container to access packages managed by pulp-rpm.

Here is a brief video demo: https://drive.google.com/file/d/1QN6yLULSZNyO_5CFvp1xEeIbN-ejrVnx/view?usp=sharing
- Unfortunately I can only share that link with Red hat emails, but feel free to reach out to me and I can share it with you

I'm not sure if you guys are interested in something like this, but thought I'd offer it up.

### DIFF
--- a/pulp_debuginfod/Containerfile
+++ b/pulp_debuginfod/Containerfile
@@ -1,0 +1,41 @@
+ARG FROM_TAG="latest"
+FROM pulp/pulp-ci-centos:${FROM_TAG}
+
+ARG PULPCORE_VERSION=""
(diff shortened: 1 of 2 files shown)


ipanova:
@pulp/installers PTAL

fao89 (approved):

PR #205, opened 2022-06-21
Avoid migration on restart

fao89:
[noissue]

### DIFF
--- a/assets/wait_on_database_migrations.sh
+++ b/assets/wait_on_database_migrations.sh
@@ -2,11 +2,18 @@
 
 database_migrated=false
 
+FILE=/database/migrated
+if test -f "$FILE"; then
+    echo "Already migrated!"
+    exit 0
+fi
+
 echo "Checking for database migrations"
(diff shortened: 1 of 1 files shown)


mdellweg:
What is this supposed to handle? I think we need to attempt to run the migrations on all startups. That is just the way we upgrade pulp in a container environment.

fao89:
the migration is only run once, at container init here:
https://github.com/pulp/pulp-oci-images/blob/latest/assets/init/000-postgres#L118
so when trying to restart services, it will be stuck in a loop:
- https://github.com/pulp/pulp-oci-images/blob/latest/assets/services/pulpcore-api/run#L3
- https://github.com/pulp/pulp-oci-images/blob/latest/assets/wait_on_database_migrations.sh#L12

mdellweg:
So in fact, this will not change whether the migrations are run, but avoid hanging in an infinite loop on container restart waiting for not running migrations to finish.

PR #208, opened 2022-06-22
Install zstd so pulp_deb can read zstd compressed packages

daviddavis:
Related to https://github.com/pulp/pulp_deb/issues/459

[noissue]


fao89 (approved):

quba42:
Starting with pulp_deb 2.19.0 (to be released tomorrow for pulpcore 3.20 compatibility), pulp_deb will require `python-debian>=0.1.44`, which is a version capable of making calls to the `unzstd` binary provided by the `zstd` package being added here. It should also be possible to backport this fix as far back as the pulp_deb 2.16 branch.

See also: https://github.com/pulp/pulp_deb/pull/513

Issue #202, opened 2022-06-16
Install podman and sudo in the CI container


dkliban:
When running `pulp_container` functional tests inside the container, `podman` is needed.

When running `pulp_rpm` functional tests inside the container, `sudo` is needed so tests don't have to differentiate between running inside a VM or a container.

PR #211, opened 2022-06-22
Install podman and sudo in the pulp_ci_centos image.

dkliban:
When running pulp_container functional tests inside the container, podman is needed.

When running pulp_rpm functional tests inside the container, sudo is needed so tests
don't have to differentiate between running inside a VM or a container.

closes: #202


dralley (approved):

fao89:
this branch isn't used in our CI

dkliban:
> this branch isn't used in our CI

on pulp_rpm 3.17 branch it is used: https://github.com/pulp/pulp_rpm/blob/3.17/.github/workflows/scripts/install.sh#L78

dkliban:
This is not needed.

PR #227, opened 2022-08-18
Exit with code 0 instead of 1 in pulpcore-api init script

### LINKED ISSUE #19: Exit the container startup if there are errors installing a plugin.
Right now if a plugin can't be installed, the environment will happily start up without installing it. This is confusing and a bad user experience.

S6 can be configured to exit if an init script fails by setting this to 2: https://github.com/pulp/oci_env/blob/main/base/compose.yaml#L38, however the pulp-api init script is designed to fail and exit if it's already been run (https://github.com/pulp/pulp-oci-images/blob/latest/assets/init/pulpcore-api#L10), which blocks the container from starting up.

Once this is fixed in the base images, the containers can be configured to exit on errors.

newswangerd:
This will make it so that the containers can be run with `S6_BEHAVIOUR_IF_STAGE2_FAILS=2`, which will cause them to shutdown if there is an error in any of the startup scripts.

Required to fix https://github.com/pulp/oci_env/issues/19

### DIFF
--- a/assets/init/pulpcore-api
+++ b/assets/init/pulpcore-api
@@ -6,7 +6,7 @@ foreground {
   }
   mkdir -p /etc/pulp/certs
 }
-if -n {
+if -n -x 0 {
   ls /etc/pulp/certs/token_private_key.pem
 }
 foreground {


newswangerd:
@dkliban this look okay to you?

fao89 (approved):

mdellweg (approved):

PR #231, opened 2022-08-19
Upgrade S6-overlay from v1 to v3

### LINKED ISSUE #230: Upgrade S6 overlay to v3

mikedep333:
from 1.22.1.0 to 3.1.2.0

fixes: #230


.github/workflows/pr_build_images.yaml
```diff
@@ -27,7 +27,12 @@ jobs:
       - uses: actions/checkout@v2
       - name: Build the images
         run: |
-          wget https://github.com/just-containers/s6-overlay/releases/download/v1.22.1.0/s6-overlay-amd64.tar.gz
+          wget https://github.com/just-containers/s6-overlay/releases/download/v3.1.1.2/s6-overlay-x86_64.tar.xz
+          wget https://github.com/just-containers/s6-overlay/releases/download/v3.1.1.2/s6-overlay-noarch.tar.xz
+          wget https://github.com/just-containers/s6-overlay/releases/download/v3.1.1.2/s6-overlay-symlinks-arch.tar.xz
+          wget https://github.com/just-containers/s6-overlay/releases/download/v3.1.1.2/s6-overlay-symlinks-noarch.tar.xz
+          unxz s6-overlay*.tar.xz
+          gzip s6-overlay*.tar
```
mdellweg, line 35: Does tar not support xz in the container?

mikedep333: No, and I'd rather save container size than install it.

pulp_ci_centos/Containerfile
```diff
@@ -2,10 +2,15 @@ FROM quay.io/centos/centos:stream8
...
+COPY s6-overlay-noarch.tar.gz /tmp
+COPY s6-overlay-symlinks-noarch.tar.gz /tmp
+COPY s6-overlay-symlinks-arch.tar.gz /tmp
 
-RUN tar xzf /tmp/s6-overlay-amd64.tar.gz -C / --exclude="./bin" && \
-    tar xzf /tmp/s6-overlay-amd64.tar.gz -C /usr ./bin
+RUN tar xzf /tmp/s6-overlay-x86_64.tar.gz -C /
+RUN tar xvf /tmp/s6-overlay-noarch.tar.gz -C /
+RUN tar xvf /tmp/s6-overlay-symlinks-noarch.tar.gz -C /
+RUN tar xvf /tmp/s6-overlay-symlinks-arch.tar.gz -C /
```
mdellweg: Would we save more space by removing those files again?

mikedep333: I think it depends on whether layers are in use or not. Either way, I'll try removing them immediately afterwards.

assets/services/certs/up
```diff
@@ -1,5 +1,9 @@
 #!/usr/bin/execlineb -P
+define PREFFIX "[oneshot] certs:"
 
+foreground {
+  if { s6-echo "${PREFFIX} ${C031}begin generating certs${C000}" }
```
mdellweg: What is the reason to run this as `if` ?

assets/test.sh
```diff
@@ -30,7 +30,7 @@ docker run --detach \
            --device /dev/fuse \
            "$image"
 sleep 10
-for _ in $(seq 20)
+for _ in $(seq 40)
```
mdellweg: Can you explain this change? Is the new s6 startup so much worse?

If the container took 10+ minutes to start up, i would consider it broken.

mikedep333: This is leftover from my earlier attempts to fix the container.

I changed it back.

dkliban (commented):
This all looks great! Can you fix the spelling of PREFFIX to PREFIX?

assets/services/certs/up
mikedep333: That's how it was done in postgres-prepare/up.

I removed it from this script, it still works.

assets/test.sh
mikedep333: I changed it to 30 to avoid a timeout on one of the 4 CI tests.

This causes it to sleep for a total of 100 seconds.

dkliban (approved):

PR #240, opened 2022-09-06
Upgrade to CentOS 9

mikedep333:
Includes postgres upgrade logic from 12 to 13

### DIFF
--- a/assets/init/postgres-init
+++ b/assets/init/postgres-init
@@ -13,16 +13,21 @@ export C035="\e[35m"
 export C036="\e[36m"
 export C037="\e[37m"
 
-PGVERSION=$(postgres --version | egrep -o "[0-9]{1,}\.[0-9]{1,}")
+PGVERSION=$(postgres --version | egrep -o "1[0-9]")
 PGHOME="/var/lib/pgsql"
 PGDATA="${PGHOME}/data"
+PGDATA_OLD="${PGHOME}/data_old"
+PGDATA_OLD_RENAMED="${PGHOME}/data_old.$(date --rfc-3339=seconds)"
 
 if [ -d "${PGDATA}/base" -a -f "${PGDATA}/PG_VERSION" ]; then
   PGDATA_VERSION=$(cat "${PGDATA}/PG_VERSION")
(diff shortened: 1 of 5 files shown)


assets/init/postgres-init
```diff
@@ -13,16 +13,21 @@ export C035="\e[35m"
 export C036="\e[36m"
 export C037="\e[37m"
 
-PGVERSION=$(postgres --version | egrep -o "[0-9]{1,}\.[0-9]{1,}")
+PGVERSION=$(postgres --version | egrep -o "1[0-9]")
```
mdellweg, line 16: Why is the old regex wrong? What is this supposed to match?

It looks quite different to me:
```
[root@f7c7ec457e7a /]# postgres --version | egrep -o "1[0-9]"
12
11
[root@f7c7ec457e7a /]# postgres --version | egrep -o "[0-9]{1,}\.[0-9]{1,}"
12.11
```

mdellweg (commented):
Very nice that we care about the upgrade now!

pulp_galaxy_ng/Containerfile
```diff
@@ -8,6 +8,6 @@ RUN pip3 install --upgrade \
   galaxy-ng${GALAXY_NG_VERSION} && \
   rm -rf /root/.cache/pip
 
-RUN ln /usr/local/lib/python3.8/site-packages/pulp_ansible/app/webserver_snippets/nginx.conf /etc/nginx/pulp/pulp_ansible.conf
```
fao89, line 11: this is the only concern I have, we'll stop to test pulp on python 3.8

assets/init/postgres-init
mikedep333: The upgrade command is based on the major version of postgresql, not the minor version anymore. The minor version was relevant for 9.x I think.

pulp_galaxy_ng/Containerfile
mikedep333: Yeah, there is no way around this. Unless we want to maintain 2 different images, which is not ideal.

I will close this PR but not delete my branch. We will implement this later.

PR #266, opened 2022-09-15
Removes all non-GA plugins from single container

### LINKED ISSUE #265: Remove all non-GA plugins from the single container
The single container ships plugins that are not GA'd currently. Specifically, the ones below. The expectation from users that I have is that the single container could be run in production so having these non GA plugins there is not aligned with that. We can add them back when they are GA, or later we can make another container that is for more experimental plugins.

```
pulp-cookbook 0.1.0b9
pulp-gem 0.0.1b2
pulp-npm 0.1.0a4
pulp-ostree 2.0.0a6
```

This change would be to remove these from the single container.

bmbouter:
This includes pulp_cookbook, pulp_gem, pulp_npm, and pulp_ostree.

closes #265

### DIFF
--- a/CHANGES/265.removal
+++ b/CHANGES/265.removal
@@ -1,0 +1,2 @@
+Removed all non-GA plugins from the single container which includes: pulp_cookbook, pulp_gem,
+pulp_npm, and pulp_ostree.
--- a/pulp_nightly/Containerfile
+++ b/pulp_nightly/Containerfile
@@ -6,13 +6,9 @@ RUN pip3 install --upgrade \
   git+https://github.com/pulp/pulp_ansible@main \
   git+https://github.com/pulp/pulp-certguard@main \
   git+https://github.com/pulp/pulp_container@main \
-  git+https://github.com/pulp/pulp_cookbook@main \
   git+https://github.com/pulp/pulp_deb@main \
   git+https://github.com/pulp/pulp_file@main \
-  git+https://github.com/pulp/pulp_gem@main \
   git+https://github.com/pulp/pulp_maven@main \
-  git+https://github.com/pulp/pulp_npm@main \
-  git+https://github.com/pulp/pulp_ostree@main \
   git+https://github.com/pulp/pulp_python@main \
   git+https://github.com/pulp/pulp_rpm@main \
   requests && \
@@ -20,5 +16,4 @@ RUN pip3 install --upgrade \
 
 RUN ln /usr/local/lib/python3.8/site-packages/pulp_ansible/app/webserver_snippets/nginx.conf /etc/nginx/pulp/pulp_ansible.conf
 RUN ln /usr/local/lib/python3.8/site-packages/pulp_container/app/webserver_snippets/nginx.conf /etc/nginx/pulp/pulp_container.conf
-RUN ln /usr/local/lib/python3.8/site-packages/pulp_cookbook/app/webserver_snippets/nginx.conf /etc/nginx/pulp/pulp_cookbook.conf
 RUN ln /usr/local/lib/python3.8/site-packages/pulp_python/app/webserver_snippets/nginx.conf /etc/nginx/pulp/pulp_python.conf


fao89 (commented):
why?

dkliban:
@bmbouter the latest branch contains only development versions of plugins. so all plugins here have a .dev version because they are installed from main branch. We probably want to change what we publish under the 'latest' tag to be the latest GA versions.

dkliban:
We should probably change this branch to publish to a 'nightly' tag. The 'latest' tag should be the latest stable releases.

bmbouter:
> why?

https://github.com/pulp/pulp-oci-images/issues/265

bmbouter:
@dkliban I'm understanding what you're saying at a high level, but I don't know how to do that. How can that be done?

fao89:
> We should probably change this branch to publish to a 'nightly' tag. The 'latest' tag should be the latest stable releases.

```
$ docker run -it ghcr.io/pulp/pulp:latest pip list | grep pulp
pulp-ansible                    0.14.1
pulp-certguard                  1.5.5
pulp-container                  2.14.0
pulp-deb                        2.19.1
pulp-file                       1.11.1
pulp-maven                      0.3.3
pulp-python                     3.7.2
pulp-rpm                        3.18.1
pulpcore                        3.21.0
```
```
$ docker run -it ghcr.io/pulp/pulp:nightly pip list | grep pulp
pulp-ansible                    0.15.0.dev0
pulp-certguard                  1.6.0.dev0
pulp-container                  2.15.0.dev0
pulp-cookbook                   0.1.0b10.dev0
pulp-deb                        2.20.0.dev0
pulp-file                       1.12.0.dev0
pulp-gem                        0.0.1b4.dev0
pulp-maven                      0.4.0.dev0
pulp-npm                        0.1.0a5.dev0
pulp-ostree                     2.0.0a7.dev0
pulp-python                     3.8.0.dev0
pulp-rpm                        3.19.0.dev0
pulpcore                        3.22.0.dev0
```

https://github.com/pulp/pulp-oci-images/blob/latest/.github/workflows/publish_images.yaml#L63-L66
https://github.com/pulp/pulp-oci-images/blob/latest/.github/workflows/publish_images.yaml#L89

bmbouter:
@fao89 so it looks like the you're saying this is already done. Is that right?

fao89:
> @fao89 so it looks like the you're saying this is already done. Is that right?

yes!

dkliban:
@bmbouter as @fao89 has pointed out, the latest tag is publishing correctly. i believe that we at least need to update https://github.com/pulp/pulp-oci-images/blob/3.21/pulp/Containerfile (3.21 branch instead of latest)

bmbouter:
That's good!

I am wondering about our labels now. From a releases perspective, any nightly is a future version of latest, so I imagined the nightly to be the same set as `latest`. This seems unusual then, we should have the nightly and the latest be the same set.

What do you think about this?

fao89:
> @bmbouter as @fao89 has pointed out, the latest tag is publishing correctly. i believe that we at least need to update https://github.com/pulp/pulp-oci-images/blob/3.21/pulp/Containerfile (3.21 branch instead of latest)

you can change it to 3.21, but pulp-ci-centos:latest is just the base image:
https://github.com/pulp/pulp-oci-images/blob/3.21/pulp_ci_centos/Containerfile

fao89:
the FROM_TAG is used to build from latest or https base image
https://github.com/pulp/pulp-oci-images/blob/latest/.github/workflows/pr_build_images.yaml#L100

the nginx from latest is different from https one:
https://github.com/pulp/pulp-oci-images/blob/latest/pulp_ci_centos/Containerfile#L108

dkliban:
@bmbouter it seems to me that in addition to the `pulp_nightly/Containerfile` you need to also modify the `pulp/Containerfile` in this PR.

dkliban:
@bmbouter you can ignore my last comment. This PR is good as it is.

bmbouter:
@fao89 @dkliban so are we good to merge this? It needs a review.

dkliban (approved):

bmbouter:
@mdellweg I'll wait to merge for you to comment since you commented on the issue too.

mdellweg:
@lubosmj will this break the pulp-cli-ostree ci?

lubosmj:
> @lubosmj will this break the pulp-cli-ostree ci?

Yes, if it affects the following piece of code:

https://github.com/pulp/pulp-cli-ostree/blob/a7706ce6f740b17067788beb066c171dfc7578d6/.ci/run_container.sh#L33
https://github.com/pulp/pulp-cli-ostree/actions/runs/2997643226/jobs/4809396786#step:8:11

bmbouter:
Thanks @lubosmj and @mdellweg. I'm reading that together we have the following goals:

* Have the tags of a container represent the same set of software
* not break the pulp-cli-ostree plugin
* Have only production software in the "pulp" container

Taken together, it sounds like we should have another container with all plugins including pre-release plugins and merging this PR. If we did that, what would that other container be called, `pulp-pre-ga-plugins`?

lubosmj:
> Taken together, it sounds like we should have another container with all plugins including pre-release plugins and merging this PR. If we did that, what would that other container be called, `pulp-pre-ga-plugins`?

The name evokes that the image will contain only pre-GA plugins even though the GA plugins will be still present there.

How about `pulp-dev-plugins`?

mdellweg:
Don't make the name too long. Why not call it `devel`?

Can we get this issue moving forward, or remove it from the board if it is stale?

bmbouter:
I'd like to see this move forward. My approach was to wait though until the PR from @mikedep333 was finished. Then we can revisit the `devel` container as a derivative of one of the container he's making. I'm not sure if this plan makes any sense or not though.

PR #276, opened 2022-09-21
Re-organize docs

fao89:
[noissue]


bmbouter (approved):
I looked through this PR and it looks great to me! The organization is much better. Thank you!

PR #277, opened 2022-09-21
Update the wait_on_database_migrations.sh script to actually wait in the face of errors

sdherr:
The `wait_on_database_migrations.sh` script would not actually wait for the database migrations if `pulpcore-manager` was returning an error for some reason. There was also a fair bit of dead code dealing with "timeouts" which never actually happen in the current version of the script, it'll wait for eternity. This PR updates it to only exit if the grep returns the "not found" exit code, and makes the output a little more correct/verbose.


fao89 (approved):
thank you!

PR #282, opened 2022-09-22
Change pulp dir owners

git-hyagi:
This is an update to try to run pulpcore containers as a regular user in k8s environments: https://github.com/pulp/pulp-operator/issues/627
[noissue]

### DIFF
--- a/images/Containerfile.core.base
+++ b/images/Containerfile.core.base
@@ -75,7 +75,9 @@ RUN ln -s /usr/lib64/libldap.so /usr/lib64/libldap_r.so
 
 RUN PULP_STATIC_ROOT=/var/lib/operator/static/ PULP_CONTENT_ORIGIN=localhost \
 	/usr/local/bin/pulpcore-manager collectstatic --clear --noinput --link && \
-	chown -R 1000 /var/lib/operator/static/
+	chown -R 700 /var/lib/operator/static/
+
+RUN chown -R 700 /var/lib/pulp/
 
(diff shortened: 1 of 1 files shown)


mikedep333:
Wow, I was literally [just working](https://github.com/pulp/pulp-oci-images/pull/283) on assigning a UID and GID to pulp in the s6 containers.

Can we please use the UID and GID 700 for Pulp?

The reason is:
1. Some users are going to use these non-s6 containers to run pulp.
2. Some of those users are going to migrate from the s6 containers to the non-s6 containers.
3. We want the UIDs and GIDs for pulp to be the same across the 2,  so as to make that migration easy without needing to chown files (or as few as possible.)
4. 1000 is a bad UID because it is the UID of the 1st user account added to the host system, and docker puts the 1000 UID directly on files on the host (it does not use subuid/subgid scheme). This also means that processes running as the 1000 UID for docker will have permissions to run commands like `sudo` very often.
5. 700 is a very safe UID/GID, in the middle of the range for system accounts (for services like flatpak, vagrant, colord, etc). Users are extremely unlikely to have 300 system accounts.

git-hyagi:
> Can we please use the UID and GID 700 for Pulp?

Sure, not a problem.
Thank you for the detailed explanation!

mikedep333 (approved):

fao89:
the galaxy images need this change:
https://github.com/pulp/pulp-oci-images/tree/latest/images/galaxy

PR #284, opened 2022-09-23
Fix CI due to plugin-template change

mikedep333:
keep_ci_update_since_branch replaced with ci_update_branches

[noissue]


mikedep333:
Note: In order to actually publish images, this requires this change:
https://github.com/pulp/pulpcore/pull/3249

fao89 (approved):

PR #311, opened 2022-10-14
Have the S6 image function as a single-process

### LINKED ISSUE #287: Layer the pulp images and make pulp/pulp for single processes
When a user starts a container using the `pulp/pulp`, the need to be able to specify a specific a service name. When a service name is provided, the container runs just that service. If a service name is not provided, the container launches s6 to manage all the services needed to run Pulp.

mikedep333:
image as well.

Implementation includes:
1. Renaming multiple images:
a. galaxy (single process) -> galaxy & galaxy-minimal
b. pulp (single process) -> pulp-minimal
c. pulp (s6) is now "pulp" on all registries
d. what was pulp / pulp-minimal with the "latest" tag is now tagged
"nightly", and "latest" is now the same as "stable"
2. Rebasing the single-process images on CentOS 8
3. Moving collect static from runtime with the s6 images to build time

fixes: #287


dkliban:
@mikedep333 We should make sure that we no longer collect static when the container starts.

dkliban (approved):
Thank you!

.github/workflows/pr_build_images.yaml
```diff
@@ -55,23 +55,65 @@ jobs:
...
+        shell: bash
+        env:
+          PY_COLORS: '1'
+      - name: Test all components (galaxy)
+        if: matrix.test_type == 'pulp_galaxy_ng'
+        run: |
+          .ci/scripts/galaxy_ng-tests.sh
+        shell: bash
+        env:
+          PY_COLORS: '1'
```
fao89, line 111: I don't think we need the compose here as it is being tested in the pulp_images.yml workflow

.github/workflows/publish_images.yaml
```diff
@@ -50,167 +43,246 @@ jobs:
...
+          podman build --file images/pulp/Containerfile --tag pulp/pulp:$VERSION .
+          podman tag pulp/pulp:$VERSION ghcr.io/pulp/pulp:$VERSION
+          podman tag pulp/pulp:$VERSION quay.io/pulp/pulp:$VERSION
+          podman tag pulp/pulp:$VERSION docker.io/pulp/pulp:$VERSION
           if [ "$VERSION" = "latest" ]; then
-            docker build --file s6_images/pulp_galaxy_ng/Containerfile --tag pulp/pulp-galaxy-ng:latest .
-            docker tag pulp/pulp-galaxy-ng:latest ghcr.io/pulp/pulp-galaxy-ng:latest
-            docker tag pulp/pulp-galaxy-ng:latest quay.io/pulp/all-in-one-galaxy:latest
+            podman tag pulp/pulp:latest ghcr.io/pulp/pulp:stable
+            podman tag pulp/pulp:latest quay.io/pulp/pulp:stable
```
fao89, line 75: 👍🏽

.github/workflows/pulp_images.yml
```diff
@@ -108,14 +187,15 @@ jobs:
           ls -al $VOLUME_PATH
           tree $VOLUME_PATH
           http --follow --timeout 30 --check-status --pretty format --print hb http://localhost:8080/pulp/api/v3/status/ || true
-  galaxy:
+      - name: Display log on error
+        if: failure()
```
fao89, line 191:
```suggestion
      - name: Display log
        if: always()
```
it is good to keep logging, if a dependency changes we can track by comparing the log from a successful run with the failed one

images/Containerfile.core.base
```diff
@@ -12,74 +16,83 @@ ENV PULP_SETTINGS=/etc/pulp/settings.py
...
 
 # Need to install optional dep, rhsm, for pulp-certguard
-RUN pip install  \
-	rhsm \
-	setproctitle \
-	gunicorn \
-        python-nginx \
-	"django-storages[boto3,azure]>=1.12.2" \
-	"requests[use_chardet_on_py3]" \
-	pulpcore
```
fao89, line 66: I would keep pulpcore, so the following images would just update some dependencies, it wouldn't need to download everything

images/Containerfile.core.base
```diff
@@ -12,74 +16,83 @@ ENV PULP_SETTINGS=/etc/pulp/settings.py
...
-			 /var/lib/pulp/media \
-			 /var/lib/pulp/assets \
-	         /var/lib/pulp/scripts \
-			 /var/lib/pulp/tmp
+RUN sed 's|^#mount_program|mount_program|g' -i /etc/containers/storage.conf
 
-# https://github.com/python-ldap/python-ldap/issues/432#issuecomment-1016756358
-RUN ln -s /usr/lib64/libldap.so /usr/lib64/libldap_r.so
+RUN groupadd -g 700 --system pulp
+RUN useradd -d /var/lib/pulp --system -u 700 -g pulp pulp
```
fao89, line 78: 👍🏽

fao89 (approved):
I just left some suggestions, but overall LGTM

.github/workflows/pr_build_images.yaml
mikedep333: But those are for different images we are testing. pr_build_images builds the big images with s6, pulp_images builds the minimal images without s6.

.github/workflows/pulp_images.yml
mikedep333: Yeah, or if the behavior is abnormal, but the tests pass.

images/Containerfile.core.base
mikedep333: I disagree.

For one thing, we may end up downgrading pulpcore later on. Because some container files let you specify the version of pulpcore.

2nd, it's a clean break between dependencies vs application.

PR #312, opened 2022-10-17
Fix gaierror

fao89:
[noissue]

### DIFF
--- a/images/assets/readyz.py
+++ b/images/assets/readyz.py
@@ -10,10 +10,14 @@ def is_api_healthy(path):
     """
     Checks if API is healthy
     """
-    address = "[::1]" if socket.has_ipv6 else "127.0.0.1"
-    url = f"http://{address}:24817{path}"
+    url = f"http://[::1]:24817{path}"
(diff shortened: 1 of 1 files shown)


images/assets/readyz.py
```diff
@@ -10,10 +10,14 @@ def is_api_healthy(path):
...
     Checks if API is healthy
     """
-    address = "[::1]" if socket.has_ipv6 else "127.0.0.1"
-    url = f"http://{address}:24817{path}"
+    url = f"http://[::1]:24817{path}"
     print(f"Readiness probe checking {url}")
-    response = requests.get(url, allow_redirects=True)
+    try:
+        response = requests.get(url, allow_redirects=True)
+    except socket.gaierror:
```
fao89, line 17: minikube doesn't support ipv6

git-hyagi (approved):
LGTM!

but I think Ina's review would be better because of my lack of knowledge in python.
/cc @ipanova 
/hold

fao89:
Any comments I can address in a follow up PR,
I'm merging to fix the CI

PR #314, opened 2022-10-17
handle ConnectionError on readiness

fao89:
[noissue]


fao89:
the lack of minikube tests, makes it hard to check if the change is really working,
merging to see if it fixes the operator CI

PR #336, opened 2022-10-28
Modify group and permissions from pulp dir

git-hyagi:
[noissue]

### DIFF
--- a/images/pulp-minimal/nightly/Containerfile.core
+++ b/images/pulp-minimal/nightly/Containerfile.core
@@ -18,3 +18,6 @@ USER pulp:pulp
 RUN PULP_STATIC_ROOT=/var/lib/operator/static/ PULP_CONTENT_ORIGIN=localhost \
        /usr/local/bin/pulpcore-manager collectstatic --clear --noinput --link
 USER root:root
+
+RUN chmod 2775 /var/lib/pulp/{scripts,media,tmp,assets}
+RUN chown :root /var/lib/pulp/{scripts,media,tmp,assets}
--- a/images/pulp-minimal/stable/Containerfile.core
+++ b/images/pulp-minimal/stable/Containerfile.core
@@ -14,3 +14,6 @@ USER pulp:pulp
 RUN PULP_STATIC_ROOT=/var/lib/operator/static/ PULP_CONTENT_ORIGIN=localhost \
        /usr/local/bin/pulpcore-manager collectstatic --clear --noinput --link
 USER root:root
+
+RUN chmod 2775 /var/lib/pulp/{scripts,media,tmp,assets}
+RUN chown :root /var/lib/pulp/{scripts,media,tmp,assets}


fao89 (approved):

mikedep333 (changes requested):
> Any RUN, CMD, ADD, COPY, or ENTRYPOINT command will be executed in the specified working directory.

https://www.educative.io/answers/what-is-the-workdir-command-in-docker

It sounds like we should add it to the other images as well. This should be in `base` then.

fao89 (approved):

images/pulp-minimal/nightly/Containerfile.core
```diff
@@ -18,3 +18,6 @@ USER pulp:pulp
 RUN PULP_STATIC_ROOT=/var/lib/operator/static/ PULP_CONTENT_ORIGIN=localhost \
        /usr/local/bin/pulpcore-manager collectstatic --clear --noinput --link
 USER root:root
+
+RUN chmod 2775 /var/lib/pulp/{scripts,media,tmp,assets}
+RUN chown :root /var/lib/pulp/{scripts,media,tmp,assets}
```
mikedep333, line 23: I don't understand this. Why are we modifying this only on the pulp-minimal images? Why do we need mode 2775? Why do we need root group ownership?

mikedep333 (changes requested):

images/pulp-minimal/nightly/Containerfile.core
git-hyagi: > "Why are we modifying this only on the pulp-minimal images?"

Why should we modify it in other images?


> "Why do we need mode 2775?"

By setting the write permission to groups + sgid we are fixing an issue where content pods were getting permission errors when trying to upload files using emptyDir.
For scenarios with persistent volumes, we are allowing users to configure the pod `securityContext.fsGroup`, which will overwrite the directories with the mount point (and will mount it with fsGroup group, not getting the permission error).

> "Why do we need root group ownership?"

If I'm not wrong, on ocp envs the container user will always be part of the root group (and another one from the supplemental group):
```
$ oc exec deployment/pulp-api -- id
uid=1000700000(1000700000) gid=0(root) groups=0(root),1000700000

$ oc get project pulp -ogo-template='{{ index .metadata.annotations "openshift.io/sa.scc.supplemental-groups"}} {{"\n"}}'
1000700000/10000

# note: this is a pod running with restricted scc
$ oc get pod/pulp-api-7664cfdfb5-sqh6r  -ogo-template='{{ index .metadata.annotations "openshift.io/scc"}} {{"\n"}}'
restricted
```

Considering that the supplemental group range will not be in the range of the gid we defined for pulp (700) and - I believe - that adding root group permission for these dirs is not a security flaw, I thought that this ownership wouldn't be a problem.


Feel free to close this PR if you know a better way to handle this situation.

mikedep333: OK, that makes sense.

However, we need to modify the other images because they can be used by the operator as well.

We should also have a comment on why we have those permissions.

Remember that we have an S6 script that modifies the permissions of the directories as well, since users will bind mount: https://github.com/pulp/pulp-oci-images/blob/latest/images/s6_assets/init/fix-pulp-perms

mikedep333: @git-hyagi we still need to modify the other containers

PR #338, opened 2022-10-29
base image: Add swig package

### LINKED ISSUE #339: Building image on arm64 fails because of missing swig
**Version**
Current git latest

**Describe the bug**

Building the pulp-minimal image fails when building the solv package from pip on arm64. This does not occur on x86_64 because there is a binary wheel for x86_64 on PyPI. Building solv from source requires the swig package which is not installed in the base image.

**To Reproduce**

```
podman build -f images/pulp-minimal/stable/Containerfile.core -t pulp/pulp-minimal:stable .
```

**Expected behavior**
An image tagged as pulp/pulp-minimal:latest

**Additional context**
Only happens on non-x86_64 platforms.

jnohlgard:
Required for building libsolv python bindings when installing from pip on non-x86_64 platforms where there is no pre-built binary wheel available from PyPI.

Fixes: #339

### DIFF
--- a/CHANGES/339.bugfix
+++ b/CHANGES/339.bugfix
@@ -1,0 +1,1 @@
+Install swig that is required when no binary wheel is available on PyPI for the solv package.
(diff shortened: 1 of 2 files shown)


mdellweg (approved):

mikedep333 (approved):

mikedep333:
@jnohlgard You need to add a file called `CHANGES/339.bugfix` that describes the change (like in your commit message.) 1 line.

jnohlgard:
Thanks, I amended the commit with a new CHANGES/339.bugfix file.

PR #344, opened 2022-10-31
The `images/compose` directory is now compatible

### LINKED ISSUE #343: Make podman-compose compatible with docker and scalable
Our compose should be compatible with docker also.

pulp-content and pulp-api should also be scalable, even at runtime, in docker (podman-compose's `up` subcommand does not support `--scale` ).

(Combining 2 features into 1 because I've already implemented it, and because 1 depends on the other.)

mikedep333:
with docker-compose in addition to podman-compose.

pulp-api and pulp-content are now also scalable,
even at runtime, in docker-compose.

The ability to access pulp-content and pulp-api at host ports 24816 and 24817 has been removed in order to implement this, access them via the nginx webserver at port 8080 instead.

fixes: #343

### DIFF
--- a/.github/workflows/pr_build_images.yaml
+++ b/.github/workflows/pr_build_images.yaml
@@ -78,8 +78,8 @@ jobs:
       - name: Compose up
         run: |
           cd images/compose
-          sed -i "s/pulp-minimal:latest/${{ matrix.test_image }}:latest/g" podman-compose.yml
-          sed -i "s/pulp\/pulp-web/quay.io\/pulp\/pulp-web/g" podman-compose.yml
+          sed -i "s/pulp-minimal:latest/${{ matrix.test_image }}:latest/g" docker-compose.yml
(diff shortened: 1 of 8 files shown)


fao89 (approved):

dkliban:
@mikedep333 does each pulpcore-api and pulpcore-content service now run alongside an nginx service?

mikedep333:
> @mikedep333 does each pulpcore-api and pulpcore-content service now run alongside an nginx service?

A single nginx server proxies to a dynamic/static list of API and a dynamic/static list of content servers.
The lists are in the DNS A records "pulp_api" and "pulp_content". On both postman and docker, the DNS record points to multiple IP addresses, the address of each content/API server. On docker at least, this list scales up and down when you "scale the service". Postman lacks a service scaling feature, so I assume it's static even if you were to ad-hoc add more servers.

dkliban (approved):

PR #351, opened 2022-11-01
Make the data dirs compatible between compose and

### LINKED ISSUE #347: As a compose user, I can transfer my settings and data between single containers and back
In order to make them easily transferable, the following limitations should be overcome with code/config changes, or at least documented:

- UID/GID on the postgres database should be 26 (compose currently uses 999, single container uses 26)
- The version of postgres used by the single container should be version 13 (compose and operator use 13, single container currently uses 12)

mikedep333:
single container
    
Implementation includes:
1. single container gets upgraded to postgres 13, and performs the
upgrade logic.
2. compose runs postgres as uid/gid 26.
3. Fixing testing in docker compose pulp-galaxy-ng with galaxy-web
4. Testing the http s6 images in single container mode, then re-using
its data with docker-compose.
5. Working around podman DNS resolution with nginx.
    
fixes: #347

### DIFF
--- a/images/s6_assets/test.sh
+++ b/images/s6_assets/test.sh
@@ -10,45 +10,62 @@ cleanup() {
 }
 trap cleanup EXIT
 
+start_container_and_wait() {
+  podman run --detach \
+             --publish 8080:$port \
+             --name pulp \
+             --volume "$(pwd)/settings":/etc/pulp:Z \
+             --volume "$(pwd)/pulp_storage":/var/lib/pulp:Z \
+             --volume "$(pwd)/pgsql":/var/lib/pgsql:Z \
+             --volume "$(pwd)/containers":/var/lib/containers:Z \
+             --device /dev/fuse \
+             -e PULP_DEFAULT_ADMIN_PASSWORD=password \
+             "$1"
+  sleep 10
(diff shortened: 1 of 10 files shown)


images/s6_assets/test.sh
```diff
@@ -10,45 +10,62 @@ cleanup() {
...
+  podman rm -f pulp
+fi
+start_container_and_wait $image
 
 if [[ ${image} != *"galaxy"* ]];then
   curl --insecure --fail $scheme://localhost:8080/assets/rest_framework/js/default.js
-  echo 127.0.0.1   pulp | sudo tee -a /etc/hosts
-  git clone --depth=1 https://github.com/pulp/pulp-cli.git
-  cd pulp-cli
+  grep "127.0.0.1   pulp" /etc/hosts || echo "127.0.0.1   pulp" | sudo tee -a /etc/hosts
```
fao89, line 60: sweet!

fao89 (approved):

images/s6_assets/test.sh
mikedep333: yeah. I'm trying to make it so you can run this script on your personal machine, without it totally clobbering the system.

PR #359, opened 2022-11-08
Fix publish_images CI

mikedep333:
[noissue]


dkliban (approved):

mikedep333:
Force merging because the only change is to a CI file that is not run for PRs.

PR #371, opened 2022-11-14
Separate the pages for Multi-Process Images vs

### LINKED ISSUE #370: Separate the Single-Process instructions from the Multi-Process instructions
As a predecessor to #332 and #333 , we should separate the pages for these 2 different sets of images.

We should also move the developer instructions (releasing instructions) to separate page.

mikedep333:
Single-Process Images

Also the developer Instructions too.

Also make many small improvements to the docs.

fixes: #370


docs/single-process-images.md
```diff
@@ -0,0 +1,80 @@
+# Single-Process Images
+
+These images are currently used on [pulp operator](https://docs.pulpproject.org/pulp_operator/), but they can be used in docker-compose or podman-compose. You can find a compose example (not production-ready) [here](https://github.com/pulp/pulp-oci-images/tree/latest/images/compose).
```
dkliban: We should not say 'not production ready' anywhere.

docs/single-process-images.md
```diff
@@ -0,0 +1,80 @@
...
+
+## pulp-minimal
+
+A single [Pulp](https://github.com/pulp/pulpcore) image that can be run as each of the following services, specified as the container command ("CMD"):
+
+- **pulp-api** - serves the Pulp(v3) API. The number of instances of this service should be scaled as demand requires.  _Administrators and users of all of the APIs put demand on this service_.
```
dkliban: The `pulp_container` registry APIs also put demand on this service. The `pulp_python` APIs also put demand on this service.

docs/single-process-images.md
mikedep333: We're literally about to delete an encryption key: https://github.com/pulp/pulp-oci-images/issues/372

We need to think of an alternate solution for warning users if we do not propagate the "not production ready" phrase (which we already say elsewhere.)

docs/single-process-images.md
mikedep333: So we should state that content consumers also put demand on this service, for those 2 plugins?

dkliban: yes, that's correct. Containers that are pushed to the registry also use this service.

docs/single-process-images.md
dkliban: I agree that we need to clearly communicate changes to consumers of these container images. We need to regularly publish a change log in our docs. We could probably also communicate about changes in a thread on discourse.

README.md
```diff
@@ -1,16 +1,15 @@
...
 
-This directory contains assets and tooling for building a variety of Pulp 3 related container images.
-
-For instructions how to run and use:
-
-- [Images](https://docs.pulpproject.org/pulp_oci_images/images/) to install Pulp 3's [first-party services](#first-party-services)
-- [S6 images](https://docs.pulpproject.org/pulp_oci_images/s6_images/) to install Pulp 3's first & [third-party services](#third-party-services)
+- [Multi-Process Images](multi-process-images) - A single image that runs [Pulp](https://github.com/pulp/pulpcore) or [Galaxy](https://github.com/ansible/galaxy_ng), as well as its third-party
+services (nginx, postgresql and redis), in one single Docker/Podman container.
+- [Single-Process Images](single-process-images) - These images are currently used on [pulp operator](https://docs.pulpproject.org/pulp_operator/), but they can be used in docker-compose or podman-compose. You can find a compose example [here](https://github.com/pulp/pulp-oci-images/tree/latest/images/compose).
```
dkliban: Maybe say used by [pulp operator]? Instead of on.

docs/multi-process-images.md
```diff
@@ -0,0 +1,112 @@
...
+
+To run all the services, you do not need to specify a container command ("CMD"). The default CMD is:
+
+- **/init** - The [s6 service manager](https://github.com/just-containers/s6-overlay) that runs all the services.
+
+The image can also function the same as the Single-Process image ["pulp-minimal"](#pulp-minimal). Specifically, the image can also be run as as each of the following individual services, specified as the container command ("CMD").
+
+- **pulp-api** - serves the Pulp(v3) API. The number of instances of this service should be scaled as demand requires.  _Administrators and users of all of the APIs put demand on this service_. If pulp_python or pulp_container are in use, _Content consumers also put demand on this service_.
+
+- **pulp-content** - serves content to clients. pulpcore-api redirects clients to pulpcore-content to download content. When content is being mirrored from a remote source, this service can download that content and stream it to the client the first time the content is requested. The number of instances of this service should be scaled as demand requires. _Content consumers put demand on this service_.
```
dkliban, line 21: pulpcore-api redirects clients to pulpcore-content in the case of pulp_ansible, pulp_python, and pulp_container. For all other plugins, the clients go directly to pulpcore-content.

mikedep333: OK, I added pulp_ansible also.

I feel that the current wording is appropriate. I think engineers/admins mostly care about which set of users will access each service.

dkliban (approved):

PR #375, opened 2022-11-15
Complete the page "Single-Process Images"

### LINKED ISSUE #332: Add a page on the `pulp-minimal` image
It should focus on usage (not building currently, assume they are pulling a built image). I imagine it would include:
* An example of running it
* The required options it needs
* Optional options it also has
* How to specify a basic pulp setting

mikedep333:
fixes: #332

Note: This is based on #371. That needs to be merged 1st, or review the latest commit only.

### DIFF
--- a/docs/single-process-images.md
+++ b/docs/single-process-images.md
@@ -12,7 +12,11 @@ A single [Pulp](https://github.com/pulp/pulpcore) image that can be run as each 
 
 - **pulp-worker** - performs syncing, importing of content, and other asynchronous operations that require resource locking. The number of instances of this service should be scaled as demand requires. _Administrators and content importers put demand on this service_.
 
-Currently built with the following plugins:
+For complete documentation on how to use this image,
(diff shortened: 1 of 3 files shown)


docs/single-process-images.md
```diff
@@ -23,6 +27,8 @@ Currently built with the following plugins:
 - [pulp_python](https://docs.pulpproject.org/pulp_python/)
 - [pulp_rpm](https://docs.pulpproject.org/pulp_rpm/)
 
+Note that the multi-process image `pulp` can also function in the same way as this image.
```
mikedep333: Maybe this should be specified in the page on the multi-process images instead?

dkliban: yeah .. it belongs in the multiprocess image docs

dkliban (changes requested):
Just the one change in line.

dkliban (approved):

PR #393, opened 2022-12-02
Add page on Migrating from pulp_installer

### LINKED ISSUE #392: Write documentation for migrating from pulp_installer to a multi-process container
See https://discourse.pulpproject.org/t/pulp-installer-3-22-will-be-the-last-release-for-the-installer/706/6

mikedep333:
to a multi-process container

fixes: #392

### DIFF
--- a/docs/migration-pulp-installer-to-multi-process.md
+++ b/docs/migration-pulp-installer-to-multi-process.md
@@ -1,0 +1,321 @@
+# Migrating from pulp_installer to a multi-process container
+
+## Overview
+
+These instructions will migrate you from a [pulp_installer deployment](https://docs.pulpproject.org/pulp_installer/) to a [multi-process container deployment](multi-process-images).
+
+The same host will be running Pulp, but in a container now.
+
+All of pulp_installer 3.23's supported distros are documented, but instructions will have extra steps depending upon the PostgreSQL version (the container runs PostgreSQL 13). These steps are all listed in the instructions per distro, but here is an overview:
+| Distro | PostgreSQL Version | Extra Steps |
+| ------ | ------------------ | ------------- |
+| Debian 11 | 13 | |
+| EL7 | 10 | Dump and restore the database |
+| EL8 | 10 | Dump and restore the database |
+| EL9 | 13 | |
+| Fedora 33 | 12 | Starting the container will take longer the 1st time |
+| Fedora 34 | 13 | |
+| Fedora 35 | 13 | |
+| Fedora 36 | 14 | PostgreSQL must remain running on the host |
+| Ubuntu 22.04 | 14 | PostgreSQL must remain running on the host |
+
+## Limitations
+
+1. All of your existing installed plugins must be installed in the multi-process container image. See the list of installed plugins [here.](multi-process-images#available-images)
+
+## Prerequisites
+
+1. Either [podman](https://podman.io/getting-started/installation) or
+[docker](https://docs.docker.com/engine/install/)/[moby-engine](https://mobyproject.org/)
+must be installed on the host.
+2. If you are running podman in rootless mode (which is recommended for security), make sure your user account has subuid's and subgid's (the default behavior for new acconts on many Linux distros.) See [this guide](https://github.com/containers/podman/blob/main/docs/tutorials/rootless_tutorial.md#etcsubuid-and-etcsubgid-configuration) to verify / set this up.
+3. If you are running docker/moby-engine, either the user account is in the docker group, or you preface `docker` commands with `sudo`.
+
+## Assumptions
+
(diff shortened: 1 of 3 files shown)


mikedep333:
These instructions are technically correct, I have tested several distros.

However, they are unclear presently. So marking draft.

docs/migration-pulp-installer-to-multi-process.md
```diff
@@ -0,0 +1,286 @@
...
+
+Run:
+```
+sudo systemctl restart postgresql
+```
+
+### Dump the PostgreSQL database (EL7 and EL8 only)
+
+```
+sudo -u postgres pg_dumpall > /tmp/dump.sql
```
dkliban, line 71: I noticed an error at this step, but it did complete successfully. 

```
[vagrant@centos8s ~]$ sudo -u postgres pg_dumpall > /tmp/dump.sql
could not change directory to "/home/vagrant": Permission denied
```

docs/migration-pulp-installer-to-multi-process.md
```diff
@@ -0,0 +1,286 @@
...
+```
+sudo systemctl disable --now nginx redis
+```
+On Ubuntu 22.04:
+```
+sudo systemctl disable --now nginx redis-server
+```
+On on EL, or Fedora prior to 36:
+```
+sudo systemctl disable --now postgresql nginx redis
```
dkliban, line 96: Why disable `postrgresql` if we configured it to listen on all interfaces and accept connections from all hosts?

docs/migration-pulp-installer-to-multi-process.md
```diff
@@ -0,0 +1,286 @@
...
+ssl
+ssl_cert_file
+ssl_key_file
+cluster_name
+stats_temp_directory
+```
+
+### Create an empty containers directory
+
+We do not bother to move `/var/lib/containers` because it is only ever used for temporary files by pulp_ansible, and may be used on the host for other purposes (like running podman):
```
dkliban: it's used by `pulp_container`, not `pulp_ansible`

docs/migration-pulp-installer-to-multi-process.md
```diff
@@ -0,0 +1,286 @@
...
+sudo mv /var/lib/pgsql pgsql
+```
+If on Debian 11:
+```
+sudo mv /var/lib/postgresql/13 pgsql
+sudo mv pgsql/main pgsql/data
+```
+If on EL7 or EL8:
+```
+sudo mv /var/lib/pgsql/data /var/lib/pgsql/data_old
```
dkliban: This moves the data to `/var/lib/pgsql/data_old`  but the following commands assume it's in `./pgsql`

docs/migration-pulp-installer-to-multi-process.md
mikedep333: Yeah, that is harmless.

docs/migration-pulp-installer-to-multi-process.md
dkliban: This should be a relative path and not absolute path.

docs/migration-pulp-installer-to-multi-process.md
```diff
@@ -0,0 +1,286 @@
...
+If running podman as root, or docker:
+```
+sudo chown -R 26:26 pgsql
+```
+
+### Configure Postgres to be compatible with the EL8-based container (Debian 11 only)
+
+Backup `pgsql/data/postgresql.conf` before you modify it:
+```
+sudo mv pgsql/data/postgresql.conf pgsql/data/postgresql.conf.old
```
dkliban: This should be a `cp`.

docs/migration-pulp-installer-to-multi-process.md
```diff
@@ -0,0 +1,286 @@
...
+
+```
+mkdir containers
+```
+
+### Restore the database (EL7 or EL8 only)
+
+Run the container with the normal [command](multi-process-images#starting-the-container), but with `-it` instead of `-detach`, and with `/bin/bash` as the specified command. 
+```
+podman run -it              --publish 8080:80              --name pulp              --volume "$(pwd)/settings":/etc/pulp:Z              --volume "$(pwd)/pulp_storage":/var/lib/pulp:Z              --volume "$(pwd)/pgsql":/var/lib/pgsql:Z              --volume "$(pwd)/containers":/var/lib/containers:Z              --device /dev/fuse              pulp/pulp /bin/bash
```
dkliban: This makes pulp available on port 8080. We should try to preserve the original port.

docs/migration-pulp-installer-to-multi-process.md
mikedep333: As we discussed, we do that on different distros.

dkliban (approved):

PR #395, opened 2022-12-08
Add gunicorn env vars (timeout and # of workers)

### LINKED ISSUE #364: As a single container user, I can specify the gunicorn timeout and number of workers
In the single container, pulpcore-api and pulpcore-content should respect the following env vars, just like the minimal container does:

PULP_GUNICORN_TIMEOUT
PULP_API_WORKERS
PULP_CONTENT_WORKERS

git-hyagi:
closes #364


mikedep333 (changes requested):
You need to modify the s6 scripts to actually use these variables.

Scripts like images/assets/pulp-content are not used by s6.

Also, single-process scripts like images/assets/pulp-content already have the variable defaulted. It's not good practice to default it twice.

fao89 (approved):

mikedep333 (approved):
Thank you!

PR #396, opened 2022-12-12
Doc signing service for multi-process container

### LINKED ISSUE #388: Write multi-process container docs for signing service
We need to document how to add the signing services to the multi-process container.

We should consider storing the scripts under /etc rather than /var/lib/pulp (to facilitate object storage eliminating /var/lib/pulp ).

git-hyagi:
closes #388

### DIFF
--- a/CHANGES/388.doc
+++ b/CHANGES/388.doc
@@ -1,0 +1,1 @@
+Added a documentation of how to add the signing services to the multi-process container.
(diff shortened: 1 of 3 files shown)


mikedep333 (approved):
Awesome, thank you!

Please write up a separate issue though about instructions for importing an existing key. It does not need to be implemented right away, users that know GPG could figure out how to do it based on these instructions. (They clearly show that it would need to be imported into the pulp user's account.)

git-hyagi:
https://github.com/pulp/pulp-oci-images/issues/398

PR #400, opened 2022-12-16
Modify container type check from probe script

### LINKED ISSUE #399: Using Pulp Operator, content-pod is never ready
**Version**
pulp-minimal:3.21.3 (currently "stable")

**Describe the bug**
When using the Pulp Operator and following the Getting Started manual (https://docs.pulpproject.org/pulp_operator/quickstart/) using the simple.yaml-sample-config to deploy, the "content"-pod never gets in ready state.

**To Reproduce**
See above.

**Expected behavior**
The "content"-pod gets in ready state, as was in pulp-minimal:3.21.2.

**Additional context**
The pod's readiness check runs the script `/usr/bin/readyz.py` which calls either 127.0.0.1:24817 or 127.0.0.1:24816 depending whether the current pod is an API-pod or a content-pod. If an environmental variable named `PULP_CONTENT_WORKERS`exists, it's a content pod. If an environmental variable named `PULP_API_WORKERS` exists, it's an API-pod. The image 3.21.3 seems to "bake" the environmental varible `PULP_API_WORKERS` into the image (see https://quay.io/repository/pulp/pulp-minimal/manifest/sha256:f3ed0ec0c60a0993a3fe6f0a6f6eee20d006faa0759dde8cd216f0aa889f4cd3), which causes the content pod to believe it's an API-pod since apparently the environmental variables are read in alphabetical order.

git-hyagi:
This commit uses the information from entrypoint, instead of checking if an env var is defined, to identify the "type" of the container. fixes #399

### DIFF
--- a/CHANGES/399.bugfix
+++ b/CHANGES/399.bugfix
@@ -1,0 +1,1 @@
+Fixed an issue with readiness probe script getting wrong pulp container type.
--- a/images/assets/readyz.py
+++ b/images/assets/readyz.py
@@ -42,8 +42,22 @@ def is_content_healthy(path):
     print("Readiness probe: ready!")
     sys.exit(0)
 
-if os.getenv("PULP_API_WORKERS"):
+
+"""
+Get container type based on entrypoint.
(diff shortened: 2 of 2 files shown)


mikedep333 (approved):
Awesome, great idea!

PR #413, opened 2023-01-13
Reassign user pulp as owner of the DB

mdellweg:
It seems that on container restart the ownership of the database may have been wrong. This leads to the unability to create trusted Postgres extensions.

[noissue]

### DIFF
--- a/CHANGES/414.bugfix
+++ b/CHANGES/414.bugfix
@@ -1,0 +1,1 @@
+Added a command to reown the db to pulp on each startup.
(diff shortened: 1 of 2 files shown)


ipanova:
@mdellweg  this fix seems related to this issue https://github.com/pulp/pulp-oci-images/issues/414  can in that case the issue be transffered here and referenced?

mdellweg:
Now that i see it working, i believe yes, we should.

mikedep333 (approved):

PR #435, opened 2023-02-14
Merge pr_build_images, publish_images and pulp_images.yml

### LINKED ISSUE #323: Publish both "x.y" and "x.y.z" tags for both sets of images
Currently, single process images have x.y.z tags, but no x.y tags.

In contrast, s6 images have x.y tags, but no x.y.z tags.

We should publish both for each.

This should be implemented after #287 is implemented.

### LINKED ISSUE #353: publish versioned s6 galaxy images
The pulp-galaxy-ng images should be published with actual version numbers of galaxy-ng.

publish_images.yml should also be separated into separate jobs for regular pulp vs galaxy-ng.

### LINKED ISSUE #361: Merge pr_build_images.yaml, publish_images.yaml and pulp_images.yml
These 2 files under .github/workflows/ have duplicated code, but only 1 gets run during pull_request CI, and are therefore increasing the number of errors.

mikedep333:
Publish both x.y branch tags and x.y.z version tags for all images. Previously, multi-process images had x.y, and single-process images had x.y.z.

Publish multi-process galaxy images with x.z branch tags and x.y.z version tags. These are the branches/versions of galaxy, not pulp.

fixes: #361

fixes: #323

fixes: #353

TODO manually afterwards: 
* Podman push x.y tags for branches for pulp-web and galaxy-web (so that the web container is available for tests that must download it.)
* Test (and review output for) non-latest branches, in particular so that the "nightly" job does not run for them.
* File an issue to improve the podman save/load process (per comments in pulp_images.yml).


.ci/scripts/update_ci_branches.py
```diff
@@ -10,9 +10,12 @@
 branches = safe_load(config)["ci_update_branches"]
 
 github_api = "https://api.github.com"
-workflow_path = "/actions/workflows/publish_images.yaml/dispatches"
-url = f"{github_api}/repos/pulp/pulp-oci-images{workflow_path}"
 
 for branch in branches:
     print(f"Updating {branch}")
+    if ( branch >= 3.22 ):
```
mikedep333: @dkliban I just realized I need to handle "latest". If I were to reverse it and check if branch < 3.22, would that cause "latest" to be treated like "3.22"?

.github/workflows/pulp_images.yml
```diff
@@ -11,14 +14,115 @@ env:
...
+            s6: true
+          - image_name: galaxy-minimal
+            web_image: galaxy-web
+            pip_name: galaxy-ng
+            s6: false
+          - image_name: galaxy
+            web_image: galaxy-web
+            pip_name: galaxy-ng
+            s6: true
+    # We do not want to build nightly images unless it's a PR, or it's the latest branch
```
mikedep333: I need to move this comment, and explain it better, up to "name: Set the list of image_variants for later jobs"

.github/workflows/pulp_images.yml
```diff
@@ -11,14 +14,115 @@ env:
...
+      # We should look into whether its possible to export just pulp-ci-centos,
+      # and tag the base image manually.
+      - name: Save podman images to tarball
+        run: podman save -m -o base-images.tar pulp/base:${TEMP_TAG} pulp/pulp-ci-centos:${TEMP_TAG}
+      - name: Cache podman images
+        uses: actions/cache/save@v3
+        with:
+          key: base-images
+          path: base-images.tar
+          fail-on-cache-miss: true
```
mikedep333, line 101: We actually want this job to fail if the step fails to upload. Because we do not want an old (by default, caches are preserved for upto 1 week) cache to be used.

Actually, the correct solution is to use a different cache key per CI run. [Instructions for this are not provided](https://github.com/actions/cache#creating-a-cache-key), but I can probably figure it out without too much effort.

.ci/scripts/update_ci_branches.py
mikedep333: Never mind, I know how to fix this now.

.github/workflows/pulp_images.yml
```diff
@@ -28,42 +133,100 @@ jobs:
...
+        if: matrix.app.s6
+        run: |
+          # Delete the database from the previous test 1st, so test.sh does not fail
+          sudo rm -rf pgsql/data
+          images/s6_assets/test.sh "pulp/${{  matrix.app.image_name }}:${TEMP_TAG}" https
+          podman stop pulp
+          podman rm pulp
       - name: Compose up
         run: |
+          # FIXME: Remove this line after finishing debuggin
```
dkliban: Do you still want to remove the set -x?

.github/workflows/pulp_images.yml
```diff
@@ -94,107 +260,59 @@ jobs:
...
         run: |
-          podman tag pulp/base:latest quay.io/pulp/base:latest
-          podman tag pulp/${APP}-minimal:$IMAGE_TAG quay.io/pulp/${APP}-minimal:$IMAGE_TAG
-          podman tag pulp/${APP}-web:$IMAGE_TAG quay.io/pulp/${APP}-web:$IMAGE_TAG
+          for registry in gchr.io docker.io quay.io; do
 
-          podman push quay.io/pulp/base:latest
-          podman push quay.io/pulp/${APP}-minimal:$IMAGE_TAG
-          podman push quay.io/pulp/${APP}-web:$IMAGE_TAG
+            # Technically multiple jobs will build and push these 2 images
```
dkliban, line 279: wouldn't it be better to cache the images and then have another job restore them and push them once?

mikedep333: I believe it would make the overall release process (and nightly runs) take longer.

Because if we entirely separate the build jobs from the push jobs, then we have to restore/load the images repeatedly. The restore and load tasks (for the base images alone) take about 1 minute each, but restore/load the app images will take even longer.

Alternatively, we could do a single job to push the base images after the unified jobs to build/push the app images. This would probably be the best approach, it would add about 1 minute to runtime.

I am leaning towards the latter.

mikedep333: Oh, also. Both would be very convoluted with the current set of tags. Because we rely on the app images to determine the 2 tags (x.y and x.y.z) for the base images. We could reduce which tags we publish for the base images, we could just use the tag of the branch. So currently, the tags would be "latest" (not "3.22"), "3.21", and older.

mikedep333: Per our convo, I separated the steps to push the base images vs push the app images. We will see how long they take.

dkliban (approved):

PR #452, opened 2023-03-02
Fix dispatching to old branches

mikedep333:
[noissue]

### DIFF
--- a/.ci/scripts/update_ci_branches.py
+++ b/.ci/scripts/update_ci_branches.py
@@ -2,6 +2,13 @@ import os
 import requests
 from yaml import safe_load
 
+def isfloat(num):
+    try:
+        float(num)
+        return True
+    except ValueError:
+        return False
+
 headers = {
     "Authorization": f"Bearer {os.environ['GITHUB_TOKEN']}",
     "Accept": "application/vnd.github.v3+json",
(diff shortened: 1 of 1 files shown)


dkliban (approved):

.ci/scripts/update_ci_branches.py
```diff
@@ -13,7 +20,7 @@
 
 for branch in branches:
     print(f"Updating {branch}")
-    if type(branch) == float and branch < 3.23 :
+    if isfloat(branch) and float(branch) < 3.23 :
```
mdellweg, line 23: Oh, so this is yaml turning `1.2` into floats, but `1.2.3` are implicitly strings? Now when some one declares a version rightfully as `"2.3"` we get a string, but your `isfloat` is still able to cast and returns `True`. Right?

Can we `pkg_resource.version.parse` to compare them? Also Why should we cap this at 3.23? All branches i see configured are less than that, and once we add 3.23 to pulpcore important branches we want to publish images, no?

mdellweg: BTW the language code for Norway equals false. No good, Yaml!

mikedep333: Hey, so when the python code parses the yaml, it interprets each branch as a string.

We verify that we can cast the string to a float.

We don't need to worry about version strings like `1.2.3`, because these are branch names only.

We cap this at 3.23 because the latest branch (which is about to become 3.23) has the CI merge. It publishes multi-process images via pulp_images.yml.

For older branches, multi-process images are published via publish_images.yaml.

mdellweg: Thanks for the clarification. But the float comparison is still not what we want.
`"3.123" > "3.23"` while `3.123 < 3.23`.
Versions are really not like numbers.

mikedep333: @mdellweg right, good point. I will fix this in a separate PR.

mikedep333: https://github.com/pulp/pulp-oci-images/pull/455

PR #455, opened 2023-03-06
Fix potential issues with version check for dispatching

mikedep333:
[noissue]

### DIFF
--- a/.ci/scripts/update_ci_branches.py
+++ b/.ci/scripts/update_ci_branches.py
@@ -1,7 +1,9 @@
 import os
 import requests
 from yaml import safe_load
+from packaging import version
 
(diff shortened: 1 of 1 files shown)


.ci/scripts/update_ci_branches.py
```diff
@@ -1,7 +1,9 @@
 import os
 import requests
 from yaml import safe_load
+from packaging import version
 
+# We use this to check if the version is x,y, like a normal branch.
```
mdellweg, line 6: Still hacky... But hey, ok the comment kind of rectifies that.

mdellweg (approved):
Thank you.
Now we compare oranges as oranges and not oranges as apples. ;)

PR #469, opened 2023-04-05
Add a bunch of needed dependencies to allow OpenTelemetry work

decko:
[noissue]

### DIFF
--- a/images/pulp_ci_centos/Containerfile
+++ b/images/pulp_ci_centos/Containerfile
@@ -31,7 +31,18 @@ RUN dnf -y module enable postgresql:13 && \
     dnf -y install redis && \
     dnf clean all
 
-RUN pip install opentelemetry-sdk opentelemetry-instrumentation-wsgi opentelemetry-distro[otlp]
+RUN pip install \
+opentelemetry-sdk \
+opentelemetry-instrumentation \
+opentelemetry-instrumentation-dbapi \
+opentelemetry-instrumentation-logging \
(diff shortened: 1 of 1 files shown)


decko:
Asked for some folks to review it just to get opinions about it, not necessarily it needs all three approvals to be accepted and merged.

images/pulp_ci_centos/Containerfile
```diff
@@ -31,7 +31,20 @@ RUN dnf -y module enable postgresql:13 && \
...
+opentelemetry-instrumentation-logging \
+opentelemetry-instrumentation-urllib \
+opentelemetry-instrumentation-wsgi \
+opentelemetry-instrumentation-aiohttp-client \
+opentelemetry-instrumentation-botocore \
+opentelemetry-instrumentation-django \
+opentelemetry-instrumentation-psycopg2 \
+opentelemetry-instrumentation-redis \
+opentelemetry-instrumentation-requests \
+opentelemetry-distro[otlp]
```
dralley, line 45: Do we need all of these?  e.g. urllib, logging and requests in particular

decko: I was not so sure if we use `requests` or not, so I left it.
Make sense to remove it?

Also, I left `logging` since it seems good to obtain spans info from the logs itself, at least in this experimental stage we're working on.

ggainey (approved):

dralley (approved):

PR #477, opened 2023-05-05
Fix the postgresql database not being preserved

### LINKED ISSUE #468: The postgresql database is not being preserved for docker-compose
As part of #347, the intention was for the volume mount for docker-compose to be changed from /var/lib/postgresql/data to /var/lib/postgresql.

However, the volume /var/lib/postgresql/data still exists, but as a non-persistent volume that always goes away when `docker-compose down` is run. This is because it is defined in the dockerfile for postgres:13 `docker inspect postgres:13`

And that directory contains the actual database.

mikedep333:
for docker-compose.

fixes: #468

Also fix the transfer process.

And fix a CI issue with podman-compose by limiting us to 1.0.3 and its healthcheck syntax.

### DIFF
--- a/.github/workflows/pulp_images.yml
+++ b/.github/workflows/pulp_images.yml
@@ -254,8 +254,12 @@ jobs:
       - name: Compose up
         run: |
           if [[ "${{ matrix.app.image_name }}" == "pulp" || "${{ matrix.app.image_name }}" == "galaxy" ]]; then
+
             # Reuse the folders from the s6 mode tests
             FILE="compose.folders.yml"
+            echo "host all all 10.0.0.0/8 trust" | sudo tee -a pgsql/data/pg_hba.conf > /dev/null
+            echo "listen_addresses = '*'" | sudo tee -a pgsql/data/postgresql.conf > /dev/null
+
             # We'll pull the web image from a registry since we didn't build it.
             if [ "${{ matrix.image_variant }}" == "nightly" ]; then
               WEB_TAG="nightly"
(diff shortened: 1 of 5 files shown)


mikedep333:
To fix existing systems:
```
curl https://raw.githubusercontent.com/alaub81/backup_docker_scripts/main/backup-docker-volume.sh -o backup-docker.volume.sh
sudo docker inspect compose_postgres_1
# Observe what the long random hexadecimal Name of the ephemeral volume is, under Mounts.
# Optional: Edit backup-docker.volume.sh so that VOLUME is only set to the specific volume Name
sudo docker-compose stop
sudo bash backup-docker.volume.sh
sudo ls -latr /backup/
# Verify that there is a tarball  with the Name of the volume, with a timestamp hyphenated after it.
sudo docker-compose down
# This next command creates the correct volume pg_datadev. The filepath must be specified
sudo docker run --rm -v /backup/volumes:/backup -v pg_datadev:/data --name restore alpine sh -c "cd /data && /bin/tar -xvf /backup/c532edb6e667574d73b6f6efc75f2e2bc39f1cce0bff3b417a6b1fec35e39b87-202305051806.tar.gz"
sudo docker-compose up
```

bmbouter (approved):
Thanks, please merge once CI passes

images/compose/docker-compose.yml
```diff
@@ -11,7 +11,7 @@ services:
       POSTGRES_INITDB_ARGS: '--auth-host=scram-sha-256'
       POSTGRES_HOST_AUTH_METHOD: 'scram-sha-256'
     volumes:
-      - "pg_data:/var/lib/postgresql"
+      - "pg_data:/var/lib/postgresql/data"
```
mikedep333, line 14: Re-Add the Z for SELinux.

dkliban (approved):

PR #483, opened 2023-05-22
Add a new Pulp image using RHEL Universal Base Image

decko:
[noissue]

### DIFF
--- a/images/pulp-rpm-ostree-ubi/Containerfile
+++ b/images/pulp-rpm-ostree-ubi/Containerfile
@@ -1,0 +1,128 @@
+FROM registry.access.redhat.com/ubi8/ubi
+
+ENV PYTHONUNBUFFERED=0
+ENV DJANGO_SETTINGS_MODULE=pulpcore.app.settings
+ENV PULP_SETTINGS=/etc/pulp/settings.py
+ENV _BUILDAH_STARTED_IN_USERNS=""
+ENV BUILDAH_ISOLATION=chroot
+ENV PULP_GUNICORN_TIMEOUT=${PULP_GUNICORN_TIMEOUT:-90}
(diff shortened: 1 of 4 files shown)


bmbouter:
I think we need an image name that identifies the plugin in it (since we'll have several that are pulp ubi based). Here are two ideas:

* pulp-rpm-ubi
* pulp-content-sources
* your idea here

bmbouter:
It's unclear to me if we'll be adding in pulp_ostree or not into this image, so maybe calling it `pulp-content-sources` would be good? Or if we do make it `pulp-rpm-ubi` we can later rename it to `pulp-rpm-ostree-ubi` would be fine too.

decko:
> I think we need an image name that identifies the plugin in it (since we'll have several that are pulp ubi based). Here are two ideas:
> 
>     * pulp-rpm-ubi
> 
>     * pulp-content-sources
> 
>     * your idea here

What about `pulp-content-source-{plugin_name}`. It would sound strange if we use `pulp-content-source-rpm-ostree`???

bmbouter:
What about `pulp-rpm-ubi` and we can change the name later?

dkliban (approved):

dralley:
@bmbouter @decko Do we have any plans to support `ubi-minimal` as an install base?

bmbouter:
> @bmbouter @decko Do we have any plans to support `ubi-minimal` as an install base?

There isn't a plan to build+publish these images as official pulp images, but the container file sources will be available in the source tree only if users want to adapt/adopt those. I don't perceive any community motivation to have a different base OS than centos, but if users wanted that we could consider building/supporting it.

bmbouter:
BTW, this was actually moved here https://github.com/pulp/pulp-clowder-deployments/pull/5

decko:
Closed since we moved it to `pulp/pulp-clowder-deployments`.

PR #485, opened 2023-05-23
Bump nginx to version 1.22

### LINKED ISSUE #484: nginx 1.14 is end-of-life and has security vulnerabilities
In the multi-process container, the EL8 appstream for nginx 1.14 went EOL in May 2021, and now has at least 1 vulnerability that a user has reported.

We should switch to nginx 1.22, the latest, supported through Nov 2025. Backup plan would be 1.20, supported through Nov 2023:
https://access.redhat.com/support/policy/updates/rhel-app-streams-life-cycle

decko:
Closes #484

### DIFF
--- a/CHANGES/484.bugfix
+++ b/CHANGES/484.bugfix
@@ -1,0 +1,1 @@
+Nginx updated to 1.22 given the EOL for 1.14.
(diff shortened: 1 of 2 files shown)


bmbouter (approved):
With only nightly failures (and those are being fixed due to non-compatible releases) I think we are ok to mere this.

PR #486, opened 2023-05-25
Remove pulp-cookbook from nightly images.

### LINKED ISSUE #490: Remove pulp-cookbook given incompatiblities with pulpcore > 3.25
Removing the pulp-cookbook plugin given it's incompatible with the new pulpcore > 3.25.

decko:
Closes #490

### DIFF
--- a/CHANGES/490.removal
+++ b/CHANGES/490.removal
@@ -1,0 +1,1 @@
+Removed ``pulp-cookbook`` from the nighlty images since it's currently incompatible with ``pulpcore``>=3.25.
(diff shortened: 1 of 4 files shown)


mikedep333 (approved):
We've never been in a situation where we removed a plugin before, but this is the nightly image, which few users probably use.

I see that pulp_cookbook doesn't work with pulpcore 3.25, so that is why we are removing it.

We have yet to publish any actual changelog. However, add a .removal notice.

I'm marking approve because the code other than the lack of .removal is fine.

Thank you!

decko:
@mikedep333 I added the CHANGE removal note there. Thanks for the suggestion.

gerrod3 (approved):

PR #487, opened 2023-05-25
Fix oneshots services being called twice

### LINKED ISSUE #474: Oneshots get called twice
**Describe the bug**
All of the oneshot scripts get called twice. To see an example of this check out the CI from galaxy_ng. postgres-prepare gets called here https://github.com/ansible/galaxy_ng/actions/runs/4765015133/jobs/8470280735?pr=1689#step:12:789 (line 788) and here https://github.com/ansible/galaxy_ng/actions/runs/4765015133/jobs/8470280735?pr=1689#step:12:1108 (line 1107).

This behavior happens for all of the oneshot scripts, which makes the containers take longer to start up.

**To Reproduce**
Run the CI centos image and search the log output from the container for `Calling /etc/init/postgres-prepare`. It will show up twice per container start.

**Expected behavior**
Oneshot scripts should only execute once.

newswangerd:
Removes `base` as a dependency from oneshot services.

closes #474

### DIFF
--- a/CHANGES/474.bugfix
+++ b/CHANGES/474.bugfix
@@ -1,0 +1,1 @@
+Fix bug causing oneshot services to run twice.


newswangerd:
@mikedep333 wdyt about this? I found your issue on s6 (https://github.com/just-containers/s6-overlay/issues/476). I'm not an expert, but I think it should be okay to have the one shots not depend on base because it looks to me like the fix-attrs service only affects longrun services, and those don't seem to get launched twice like the oneshots do.

mikedep333:
Hi @newswangerd ,

If this works, great.

However, please test my original theory from the bug. Try launching with `-e PULP_WORKERS=1`. If the oneshots still run, then we can merge. Thank you!

newswangerd:
@mikedep333 did you have time to test it?

>

newswangerd:
> Hi @newswangerd ,
> 
> If this works, great.
> 
> However, please test my original theory from the bug. Try launching with `-e PULP_WORKERS=1`. If the oneshots still run, then we can merge. Thank you!

Did you have time to test this?

PR #489, opened 2023-05-26
Removes the OpenTelemetry installation from the `pulp-ci-centos` image.

### LINKED ISSUE #488: Remove OpenTelemetry dependencies from pulp-ci-centos image
Since we've added the OpenTelemetry libraries to `pulpcore` itself, we should remove it from the `pulp-ci-centos` image as well.

decko:
Closes #488


bmbouter (approved):
Now that these deps are in pulpcore I think we should remove them from here. Thanks!

PR #498, opened 2023-06-07
Add a new Pulp UBI based image with pulp-gem and pulp-certguard.

decko:
Adds a new pulp ubi based image with just gem plugin.

[noissue]


decko:
Closed since we're moving it to `pulp/pulp-gem-clowder-deployment`

PR #501, opened 2023-06-09
Includes `pulp_ostree` as one of the plugins installed in the images.

### LINKED ISSUE #491: Mention that pulp_ostree is now a part of the one-in-all container in the documentation

decko:
Closes #491


bmbouter (approved):
I git grepped and these are the only entries I saw for plugins. Thanks!

PR #508, opened 2023-06-15
Fix CI failing to compose-up for 2 reasons

mikedep333:
Fix CI failing to compose-up the nightly s6 images

because it is looking for the non-existing tags of the development branch's web images. It should use "nightly" instead.

[noissue]

 Workaround CI failing due to an ubuntu podman bug
    
 https://bugs.launchpad.net/ubuntu/+source/libpod/+bug/2024394
    
 Workaround will disable itself when no longer needed
    
 [noissue]

### DIFF
--- a/.github/workflows/pulp_images.yml
+++ b/.github/workflows/pulp_images.yml
@@ -242,7 +242,11 @@ jobs:
             # Reuse the folders from the s6 mode tests
             FILE="docker-compose.folders.yml"
             # We'll pull the web image from a registry since we didn't build it.
-            WEB_TAG="${APP_BRANCH}"
+            if [ "${{ matrix.image_variant }}" == "nightly" ]; then
+              WEB_TAG="nightly"
+            else
+              WEB_TAG="${APP_BRANCH}"
+            fi
           else
             FILE="docker-compose.yml"
             WEB_TAG="${TEMP_APP_TAG}"
--- a/.github/workflows/pulp_images.yml
+++ b/.github/workflows/pulp_images.yml
@@ -186,6 +186,12 @@ jobs:
           echo ::endgroup::
           echo "HTTPIE_CONFIG_DIR=$GITHUB_WORKSPACE/.ci/assets/httpie/" >> $GITHUB_ENV
           pip install podman-compose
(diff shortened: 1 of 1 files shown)


decko (approved):

.github/workflows/pulp_images.yml
```diff
@@ -242,7 +242,10 @@ jobs:
             # Reuse the folders from the s6 mode tests
             FILE="docker-compose.folders.yml"
             # We'll pull the web image from a registry since we didn't build it.
-            WEB_TAG="${APP_BRANCH}"
+            if [ "${{ matrix.image_variant }}" == "nightly" ]
+              WEB_TAG="nightly"
+            else
+              WEB_TAG="${APP_BRANCH}"
```
decko: I think we need to close this `if` block here.
```suggestion
            if [ "${{ matrix.image_variant }}" == "nightly" ] then
              WEB_TAG="nightly"
            else
              WEB_TAG="${APP_BRANCH}"
            fi
```

decko (changes requested):

.github/workflows/pulp_images.yml
```diff
@@ -242,7 +242,11 @@ jobs:
             # Reuse the folders from the s6 mode tests
             FILE="docker-compose.folders.yml"
             # We'll pull the web image from a registry since we didn't build it.
-            WEB_TAG="${APP_BRANCH}"
+            if [ "${{ matrix.image_variant }}" == "nightly" ]
```
decko: I think we're missing a `then` by the end of this line, like this:
```suggestion
            if [ "${{ matrix.image_variant }}" == "nightly" ] then
```

decko (changes requested):

.github/workflows/pulp_images.yml
mikedep333: You forgot the semicolon.

Fixed it :)

mikedep333:
Looks like the 2 remaining CI failures are due to an additional issue.

I'd like to merge, and then fix this next issue.

decko (approved):

PR #525, opened 2023-07-27
Fix CI

gerrod3:
Not entirely sure why so many of the runners are failing, but it is preventing the images from being properly built and published. Going to try to fix them all in this PR, one force-push at a time.


images/s6_assets/test.sh
```diff
@@ -69,10 +69,12 @@ if [[ ${image} != *"galaxy"* ]];then
   if [ -d pulp-cli ]; then
```
mdellweg: It sounds strange that we do not even know whether we have the cli installed at this time.
Can we just do the same as we do in the plugin_ci and use a released version of the CLI?

gerrod3: I think it is because this script is ran multiple times during the CI, thus the need to check.

gerrod3:
Ok, I think I got it all fixed. For the galaxy CIs they were breaking because of the pip dependency resolution failing on opentelemetry's buggy mess. I  simply flipped the order of the pip install around so that galaxy is downloaded first which will pull the correct pulpcore since pulpcore is a dependency of galaxy. We will probably still run into problems like these in the future, but I can't solve that now.

The pulp CI was fixed in the CLI (https://github.com/pulp/pulp-cli/pull/753). I still have it set to pull from upstream CLI so still at risk of bad CLI changes breaking everything. Should I have it just install the latest release instead?

Finally I also refactored a bit of the workflow to make it faster (and easier to read). Two of the steps were never being ran, so I deleted them, but took their scripts and incorporated them into `images/s6_asset/test.sh`. I also deleted the retest of the pulp image after the upgrade tests since they were both testing the same version (this should make it ~20 mins faster).

mdellweg (approved):
I'm sure there is room for more improvement, but if this is what it takes to unblock us...

.github/workflows/pulp_images.yml
```diff
@@ -128,19 +140,15 @@ jobs:
           - image_name: pulp-minimal
             web_image: pulp-web
             pip_name: pulpcore
-            s6: false
```
mikedep333, line 131: If you think this improves how easy it is to understand pulp_images.yml, I am OK with this.

.github/workflows/pulp_images.yml
```diff
@@ -225,30 +242,25 @@ jobs:
           echo "BASE_BRANCH: ${base_branch}"
           echo "BASE_VERSION=${base_version}" >> $GITHUB_ENV
           echo "BASE_BRANCH=${base_branch}" >> $GITHUB_ENV
-      - name: Test upgrading to the image in s6 mode
+
+      - name: Test image with upgrade in s6 mode (pulp)
```
mikedep333, line 246: This is a good renaming.

images/galaxy-minimal/stable/Containerfile.core
```diff
@@ -5,8 +5,8 @@ ARG PULPCORE_VERSION=""
 ARG GALAXY_NG_VERSION=""
 
 RUN pip3 install --upgrade \
-  pulpcore${PULPCORE_VERSION} \
-  galaxy-ng${GALAXY_NG_VERSION} && \
+  galaxy-ng${GALAXY_NG_VERSION} \
+  pulpcore${PULPCORE_VERSION} && \
```
mikedep333, line 9: I see you switched the order around. Does this fix something?

images/galaxy/nightly/Containerfile
```diff
@@ -4,8 +4,8 @@ FROM pulp/pulp-ci-centos:${FROM_TAG}
 ARG GALAXY_NG_VERSION=""
 
 RUN pip3 install --upgrade \
-  requests \
-  git+https://github.com/ansible/galaxy_ng.git@master && \
+  git+https://github.com/ansible/galaxy_ng.git@master \
```
mikedep333, line 7: I see you switched the order around. Does this fix something?

images/s6_assets/galaxy_ng_tests.sh
```diff
@@ -3,30 +3,30 @@ set -euo pipefail
...
+BASE_ADDR="$scheme://$SERVER:$WEB_PORT"
 echo "Base Address: $BASE_ADDR"
 REPOS=( "published" "staging" "rejected" "community" "rh-certified" )
 REPO_RESULTS=()
 
 echo "Waiting ..."
 sleep 10
 
-TOKEN=$(curl --location --request POST "$BASE_ADDR/api/galaxy/v3/auth/token/" --header 'Authorization: Basic YWRtaW46cGFzc3dvcmQ=' --silent | python3 -c "import sys, json; print(json.load(sys.stdin)['token'])")
+TOKEN=$(curl --insecure --location --request POST "$BASE_ADDR/api/galaxy/v3/auth/token/" --header 'Authorization: Basic YWRtaW46cGFzc3dvcmQ=' --silent | jq -r .token)
```
mikedep333, line 18: Nice simplification!

images/s6_assets/galaxy_ng_tests.sh
```diff
@@ -3,30 +3,30 @@ set -euo pipefail
 
 SERVER="localhost"
 WEB_PORT="8080"
+scheme=${1:-http}
 
 pip3 install "ansible<2.13.2"
 
-BASE_ADDR="http://$SERVER:$WEB_PORT"
-echo $BASE_ADDR
```
mikedep333, line 10: This line would be nice to have for debugging, albeit with a better print statement:
```suggestion
echo "BASE_ADDR: $BASE_ADDR"
```

mikedep333 (approved):
Thank you very much! A few optional comments.

images/galaxy-minimal/stable/Containerfile.core
gerrod3: Yeah, `pip` dependency resolver somehow fails, even when it shouldn't, when pulpcore is listed first. Since pulpcore is a dependency of galaxy it makes sense to let `pip` try to install galaxy first and then see if the requested pulpcore is compatible.

images/s6_assets/galaxy_ng_tests.sh
gerrod3: I deleted that line since it is a duplicate, the line right below prints the same thing with the better message.

PR #529, opened 2023-08-01
Fix 3.23 version pin

gerrod3:
Now the 3.23 won't publish the latest y-stream.


mdellweg (approved):
Darn! I always for get how this shorthand works.

mdellweg (approved):
pip weirdness?

gerrod3:
> pip weirdness?

Yep, same issue that I recently just fixed. I just manually backported it to this PR.

PR #538, opened 2023-08-29
Use django to check db connection

mdellweg:
[noissue]

### DIFF
--- a/images/assets/wait_on_postgres.py
+++ b/images/assets/wait_on_postgres.py
@@ -1,40 +1,22 @@
 #!/usr/bin/env python3
 
-import os
-import socket
 import sys
 import time
+from django.db import connection, utils
 
 if __name__ == "__main__":
 
-    postgres_is_alive = False
-    s4 = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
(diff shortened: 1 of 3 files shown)


images/assets/wait_on_postgres.py
```diff
@@ -1,40 +1,22 @@
 #!/usr/bin/env python3
 
-import os
-import socket
 import sys
 import time
+from django.db import connection, utils
```
mikedep333, line 5: Extremely good idea!

images/compose/compose.folders.yml
```diff
@@ -51,8 +51,6 @@ services:
       - "./assets/certs:/etc/pulp/certs:z"
       - "../../pulp_storage:/var/lib/pulp:z"
     environment:
-      POSTGRES_SERVICE_PORT: 5432
```
mikedep333, line 54: I see these are not necessary because they're in the settings.py file. 👍🏻

mikedep333 (approved):

images/compose/compose.folders.yml
mdellweg: Actually it was only used by the old wait_on_postgres.

PR #542, opened 2023-08-31
Compose secrets

mdellweg:


images/compose/compose.folders.yml
```diff
@@ -1,30 +1,46 @@
-version: '3'
+# This file is intended as a starting point. Before using it, make a copy and adjust at least the passwords.
```
gerrod3: Shouldn't this comment make reference to the new password files? It kind of reads like the passwords should be changed here.

images/compose/compose.folders.yml
```diff
@@ -1,30 +1,46 @@
...
+---
+version: '3.7'
```
gerrod3: Is there any reason why the version is now 3.7?

images/compose/compose.folders.yml
```diff
@@ -42,16 +61,23 @@ services:
...
     depends_on:
       redis:
+        condition: service_healthy
       postgres:
         condition: service_healthy
     hostname: pulp-api
     user: pulp
     volumes:
       - "./assets/settings.py:/etc/pulp/settings.py:z"
-      - "./assets/certs:/etc/pulp/certs:z"
```
gerrod3, line 106: Why are you removing the certs directory? Isn't this crucial for setting up cert auth and the CertGuard plugin?

images/compose/compose.folders.yml
```diff
@@ -76,13 +109,31 @@ services:
...
+      - db_encryption_key
+    networks:
+      - pulp_internal
+    restart: always
+
+networks:
+  pulp_internal:
+
+secrets:
+  app_secret:
```
gerrod3: Very cool use of compose features!

images/compose/compose.yml
gerrod3, line 1: What is suppose to be the difference between this file and the compose.folders.yml file?

images/compose/compose.folders.yml
mdellweg: Not part of this pr anymore...
Also not a particular reason. Just from examples.

images/compose/compose.folders.yml
mdellweg: Cert auth, I'm not sure, but certguard will have all the certs in the database.

mdellweg: Would they auth and webserver certs not be in /etc/pki ?

mdellweg: Also tls should be terminated by nginx and not pulp-api.

images/compose/compose.yml
mdellweg: compose.folders is using directories on the host instead of volumes. That's why the diff should be very small.

mdellweg:
This is not the PR you are looking for, StaleBot!

PR #547, opened 2023-09-14
Fix container workaround with stable url

mdellweg:
[noissue]


mdellweg:
Oh!!! It took me some time to realize the last failing workflow is not this issue, but that fact that the workflow has false assumptions about the existance of a pulp/galaxy-web:4.8 image.

dkliban (approved):

PR #548, opened 2023-09-18
Add an option to skip migrations when the application is starting.

### LINKED ISSUE #544: The entrypoints for services are not compatible with kubernetes/compose
The entrypoint for pulp-api runs migrations and sets admin password. This should be done outside of this script.

decko:
Closes #544

### DIFF
--- a/CHANGES/544.feature
+++ b/CHANGES/544.feature
@@ -1,0 +1,1 @@
+Adds an option to skip migration and setting the admin password.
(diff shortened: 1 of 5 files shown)


mdellweg:
What is the plan with this? I see adding another layer of complexity on top of what is already much too complex.
The issue description was: Stop running setup code in the api container.

So the question here is: How are the pulp-minimal containers to be used?
I see:
- compose files in this repository (we control them).
- pulp_operator (we can tell it to run a setup routine).

decko:
After some discussion we decided to go on with another idea. Closing it for now.

PR #551, opened 2023-09-19
Split up some responsabilities from the entrypoint to a service container.

### LINKED ISSUE #544: The entrypoints for services are not compatible with kubernetes/compose

decko:
Closes #544

### DIFF
--- a/.github/workflows/pulp_images.yml
+++ b/.github/workflows/pulp_images.yml
@@ -17,7 +17,7 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           # by default, it uses a depth of 1
           # this fetches all history so that we can read each commit
@@ -75,7 +75,7 @@ jobs:
           echo "Building $temp_base_tag"
           echo "TEMP_BASE_TAG=${temp_base_tag}" >> $GITHUB_ENV
 
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           # by default, it uses a depth of 1
           # this fetches all history so that we can read each commit
@@ -162,7 +162,7 @@ jobs:
           echo "TEMP_APP_TAG=${temp_app_tag}" >> $GITHUB_ENV
           echo "TEMP_BASE_TAG=${temp_base_tag}" >> $GITHUB_ENV
 
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           # by default, it uses a depth of 1
           # this fetches all history so that we can read each commit
@@ -271,14 +271,14 @@ jobs:
           sed -i "s/pulp-web:latest/${{ matrix.app.web_image }}:${WEB_TAG}/g" $FILE
           id | grep "(root)" || sudo usermod -G root $(whoami)
           podman-compose -f $FILE up -d
-          sleep 30
+          podman exec compose_pulp_api_1 /usr/bin/wait_on_database_migrations.sh
           for _ in $(seq 20)
           do
-            sleep 3
             if curl --fail http://localhost:8080/pulp/api/v3/status/ > /dev/null 2>&1
             then
               break
             fi
+            sleep 3
           done
           curl --fail http://localhost:8080/pulp/api/v3/status/ | jq
         shell: bash
--- a/images/assets/pulp-api
+++ b/images/assets/pulp-api
@@ -1,44 +1,7 @@
-#!/bin/bash -x
-
-mkdir -p /var/lib/pulp/media \
-         /var/lib/pulp/assets \
-         /var/lib/pulp/tmp
+#!/bin/bash -eu
 
 /usr/bin/wait_on_postgres.py
-
-# Get list of installed plugins via pip
-# Assumes they are all named like "pulp-file" -> "file", with no 2nd dash.
-# (Was previously needed when we ran `pulpcore-manager makemigrations`)
-# PLUGINS=$(pip list | awk -F '[[:space:]]+|[-]' '/pulp-/{printf $2 " " }')
-
-/usr/local/bin/pulpcore-manager migrate --noinput
-
-set +x
-
-if [ -n "${PULP_SIGNING_KEY_FINGERPRINT}" ]; then
-    /usr/local/bin/pulpcore-manager add-signing-service "${COLLECTION_SIGNING_SERVICE}" /var/lib/pulp/scripts/collection_sign.sh "${PULP_SIGNING_KEY_FINGERPRINT}"
-    /usr/local/bin/pulpcore-manager add-signing-service "${CONTAINER_SIGNING_SERVICE}" /var/lib/pulp/scripts/container_sign.sh "${PULP_SIGNING_KEY_FINGERPRINT}" --class container:ManifestSigningService
-fi
-
-if [[ -n "$PULP_DEFAULT_ADMIN_PASSWORD" ]]
-then
-  PASSWORD_SET=$(/usr/local/bin/pulpcore-manager shell -c "from django.contrib.auth import get_user_model; print(get_user_model().objects.filter(username=\"admin\").exists())")
-  if [ "$PASSWORD_SET" = "False" ]
-  then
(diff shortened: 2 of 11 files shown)


images/assets/pulp-content
```diff
@@ -1,6 +1,7 @@
-#!/bin/bash -x
+#!/bin/bash
```
mdellweg: Have you considered using `set -eu`?

images/compose/compose.yml
```diff
@@ -56,14 +95,16 @@ services:
         condition: service_healthy
       postgres:
         condition: service_healthy
+      migration_service:
+        condition: service_completed_successfully
+      signing_key_service:
+        condition: service_completed_successfully
```
mdellweg: I fear, this won't work with podman-compose, unless we keep the containers running there.

images/pulp-minimal/stable/Containerfile.core
```diff
@@ -27,7 +27,17 @@ RUN pip3 install --upgrade \
 
 # Prevent pip-installed /usr/local/bin/pulp-content from getting run instead of
 # our /usr/bin/pulp-content script.
-RUN rm -f /usr/local/bin/pulp-content
+# RUN rm -r /usr/local/bin/pulp-content
```
mdellweg: I know this is a hack, but we need it for older pulpcore.

decko: I'm checking it, but seems I'm unable to revert it. :sob:

images/compose/compose.yml
decko: Yeah, it won't work with podman-compose for now. The idea is to enable docker-compose to use it and emulate the same behavior using the `wait_for_`* scripts.

images/assets/pulp-content
decko: Thanks for the suggestion

images/pulp-minimal/stable/Containerfile.core
mdellweg: Don't confuse `-f` and `-r` as i did.

images/compose/compose.yml
```diff
@@ -31,12 +31,51 @@ services:
...
+    depends_on:
+      postgres:
+        condition: service_healthy
+    command: pulpcore-manager migrate --noinput
+    volumes:
+      - "./assets/settings.py:/etc/pulp/settings.py:z"
+      - "./assets/certs:/etc/pulp/certs:z"
+      - "pulp:/var/lib/pulp"
+
+  # signing_key_service:
```
dkliban: Did you mean to leave the signing_key_service commented out?

decko: Noooooope. Fixing it now.

images/assets/pulp-worker
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash -x
+#!/bin/bash -xeu
```
mdellweg:
```suggestion
#!/bin/bash -eu
```

images/compose/compose.folders.yml
```diff
@@ -56,14 +93,18 @@ services:
         condition: service_healthy
       postgres:
         condition: service_healthy
+      migration_service:
+        condition: service_completed_successfully
+      set_init_password_service:
+        condition: service_completed_successfully
```
mdellweg, line 99: This is not a technical dependency. Pulp can run just fine without that password. I think we should drop these lines.

In case of a configured SSO solution, you can still use pulp without admin.

images/assets/wait_on_database_migrations.sh
```diff
@@ -2,7 +2,7 @@
 
 echo "Checking for database migrations"
 while true; do
-  /usr/local/bin/pulpcore-manager showmigrations | grep '\[ \]'
+  /usr/local/bin/pulpcore-manager showmigrations | grep '\[ \]' &> /dev/null
```
mdellweg, line 5: :+1: Really a good change.
I thought, I saw you having a django based solution to this. Is improving this script left for another day?

.github/workflows/pulp_images.yml
```diff
@@ -271,14 +271,14 @@ jobs:
           sed -i "s/pulp-web:latest/${{ matrix.app.web_image }}:${WEB_TAG}/g" $FILE
           id | grep "(root)" || sudo usermod -G root $(whoami)
           podman-compose -f $FILE up -d
-          sleep 30
+          podman exec compose_pulp_api_1 /usr/bin/wait_on_database_migrations.sh
```
mdellweg, line 274: Is this what make it work with podman-compose? I'd love to hear the reasoning.

mdellweg (approved):

images/assets/wait_on_database_migrations.sh
decko: It didn't worked out. :disappointed:
I'll open an issue for it so we can have more eyes on it.

.github/workflows/pulp_images.yml
decko: Nope. Here the thing is to wait for all the migrations to be applied. My suspicious about the way it was working before it's because we didn't used the `healthy` checker, and within 30 seconds you have the exact amount of migrations needed to make the `status` endpoint work. It would take at least 3 or 4 more minutes to finish the migrations, which are now needed to change the status of the `api` and `content` containers to `healthy`.

images/compose/compose.folders.yml
decko: I vote for leave it as it is now. Or at least leave it as comment.
It would make easier for new users to find how to change the password without using `podman exec ...`.

images/assets/pulp-worker
decko: :exploding_head:

decko: Done. And thanks for look closely.

images/compose/compose.folders.yml
mdellweg: I'm not saying we should remove the service container. We just don't need it as a requirement on the api pod.
It will still start up anyway unless you call podman-compose in a special way.

images/assets/wait_on_database_migrations.sh
mdellweg: Absolutely. We don't _need_ to do it here.

.github/workflows/pulp_images.yml
mdellweg: 3 to 4 minutes? We _need_ to talk about our upgrade strategy once more and how we can safely squash migrations.

dkliban (approved):

images/assets/pulp-api
```diff
@@ -1,44 +1,7 @@
-#!/bin/bash -x
-
-mkdir -p /var/lib/pulp/media \
-         /var/lib/pulp/assets \
-         /var/lib/pulp/tmp
```
rooftopcellist, line 5: @decko @mdellweg  It looks like the `/var/lib/pulp/tmp` directory is no longer being pre-created, which causes galaxy worker containers to CrashLoopBackOff.  Is this a problem for Pulp as well?  Should the fix be here, or in the Dockerfile?

mdellweg: /var/lib/pulp is mounted as a volume. This must happen at runtime. I wonder if the services using it are creating these dirs on demand? How does it even work for pulp these days @decko ?

decko: We have it created in the core base image. https://github.com/pulp/pulp-oci-images/blob/ccbb90e19fb4dbc20afca2d5455533db34f25526/images/Containerfile.core.base#L98

decko: > /var/lib/pulp is mounted as a volume. This must happen at runtime. I wonder if the services using it are creating these dirs on demand? How does it even work for pulp these days @decko ?

Ok, so it was being re-created after the /var/lib/pulp being mounted.
Should it have it's mount point?

mdellweg: Our compose file mounts a volume on /var/lib/pulp. So i suspect one of the setup processes is probably recreating these dirs?

decko: > Our compose file mounts a volume on /var/lib/pulp. So i suspect one of the setup processes is probably recreating these dirs?

Yes, we mount, but we're not recreating those directories during the runtime. Since this is a tmp dir, would make send for pulp itself create it if it's not found?

PR #553, opened 2023-09-27
Configure SSL if variable is capitalized or not

tinsjourney:
When using ansible-podman-collections to create the multi-process images, boolean environment are capitalized.
So this is workaround until
https://github.com/containers/ansible-podman-collections/issues/648 is fixed.

### DIFF
--- a/images/s6_assets/init/nginx
+++ b/images/s6_assets/init/nginx
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-if [ "${PULP_HTTPS}" = "true" ]; then
+if [ "${PULP_HTTPS,,}" = "true" ]; then
   cp -avr /nginx/ssl_nginx.conf /etc/nginx/nginx.conf
 else
   cp -avr /nginx/nginx.conf /etc/nginx/nginx.conf


images/s6_assets/init/nginx
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-if [ "${PULP_HTTPS}" = "true" ]; then
+if [[ "${PULP_HTTPS}" = "true" || "${PULP_HTTPS}" = "True" ]]; then
```
mdellweg:
```suggestion
if [ "${PULP_HTTPS,,}" = "true" ]; then
```
I am +1 for making the matching less strict.
But you should probably not define your environment variables with bool in the first place.
Yaml does not like Norway...

mdellweg (changes requested):

mdellweg:
Instead of closing, the PR, would you be willing to apply the suggestion and we can merge it?

tinsjourney:
@mdellweg OK I reopened the PR and did the change.

mdellweg (approved):

mdellweg:
Can you please squash to commits into one?

PR #558, opened 2023-10-24
Use the official and update nginx image.

### LINKED ISSUE #557: Update nginx to fix the HTTP2 Rapid Reset issue.
Would be great to use a fixed nginx version on any usage we have of it.
A quick search just pointed to all `minimal` images, and the `multi-process` images.

For the minimal images, maybe we should use the official `nginx` image from the docker hub.
For the multi-process image we need to check the latest nginx version available for Stream8.

decko:
Closes #557


PotentialIngenuity:
<strike> @ggainey Can we get this PR reviewed and merged? Our vulnerability scanners are flagging pulp for running an nginx version <1.17.7 </strike>

<strike> I see that #485 fixed this for the all in one but the minimal container was missed. </strike>

nvm. fixed here #703

PR #546, opened 2023-09-13
Enable multi-arch builds again

StopMotionCuber:
This is the same as for #367 and should fix #366. For some reason that commit got reverted and I'm not sure why.

PR #562, opened 2023-11-20
Fix builds for multiarch

StopMotionCuber:
This is providing fixes for multiarch builds. I've hardcoded the desired architectures at some places, this could probably be streamlined in case wanted.

I've excluded `ppc64le`, as it had various issues with dependencies for `pip install pulpcore`, of which `pip install grpcio` was not fixable. There was an additional error with pyyaml in version 6.0, but this is fixed in 6.0.1 where some version pinning while building would probably have worked.

[The GitHub Actions run](https://github.com/StopMotionCuber/pulp-oci-images/actions/runs/6930543170/) went mostly fine and only complained about the `docker login` (which obviously doesn't work without credentials)

### DIFF
--- a/.github/workflows/pulp_images.yml
+++ b/.github/workflows/pulp_images.yml
@@ -104,8 +104,11 @@ jobs:
           podman version
           buildah version
           sudo podman run --rm --privileged multiarch/qemu-user-static --reset -p yes
-          podman build --platform linux/arm64,linux/amd64,linux/ppc64le --format docker --file images/Containerfile.core.base --tag pulp/base:${TEMP_BASE_TAG} .
-          podman build --platform linux/arm64,linux/amd64,linux/ppc64le --format docker --file images/pulp_ci_centos/Containerfile --tag pulp/pulp-ci-centos:${TEMP_BASE_TAG} --build-arg FROM_TAG=${TEMP_BASE_TAG} .
+          for ARCH in arm64 amd64
+          do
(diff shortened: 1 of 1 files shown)


StopMotionCuber:
Ah, PR linting failing again cause of changelog building automation and stuff.

Would be nice if a maintainer could get back to me with the preferred way of handling that. Squash commits + open a ticket + add that to Changes/XXX.bugfix? `[noissue]` to all commits? Squash and `[noissue]` for that single commit?

lubosmj:
To unblock you, try squashing all the commits and adding "[noissue]" to the commit description:
```
git commit -m "Blabla" -m "[noissue]"
```
We can later decide on whether to open a new issue or not...

StopMotionCuber:
Done, commits are squashed now

mdellweg:
Any explanation, why the PR was green, while the nightly builds stopped working here?

StopMotionCuber:
> Any explanation, why the PR was green, while the nightly builds stopped working here?

From my side I'm not sure about that. My experience with GitHub, especially GitHub Actions and their checking is limited, have more experience with GitLab.

Especially I'm seeing only 2 passing checks on #546 while there are 11 on this PR. Not sure if the image build was not done or the history is just confusing...?
Anyway, it has probably built _something_ which is why the light turned green

hstct:
Currently I'm unable to build any other architecture than arm64 using the latest images. podman seems to ignore any options to force the platform to amd64. Hopefully this issue will fix this issue?

lubosmj (approved):

Issue #565, opened 2023-11-22
Building and publishing multi-arch images had an impact on project's CI


decko:
After getting some PRs where we build multi-arch images (amd64,arm64) our CI can't publish new images anymore, neither run the tests, like [here](https://github.com/pulp/pulpcore/actions/runs/6952326126/job/18915733298#step:9:449)

StopMotionCuber:
I guess that's on me. I've just seen that the last fix I've submitted should have not only been applied to "Push app images to registries" but also to "Push base images to registries". Unfortunately it's not that easy debugging some of that stuff without the credentials to actually login to registries and push stuff in a different environment.

Contributions still welcome? 😬

mdellweg:
At least we have some trail of information what went wrong. Sorry that we couldn't wait for it to be fixed in place. This really threw a wrench in the gearbox of our project wide CI's.
I believe we will be more careful next time and ideally we could catch these kind of issues in the PR's (not sure it's possible). I can only speak for myself: I'd still welcome the contribution.

ipanova:
> Contributions still welcome? 😬

In development things go sideways all the time. Better having iterations with mistakes that'd eventually lead to success rather than a flat line  ![waveform-path-ecg](https://github.com/pulp/pulp-oci-images/assets/4253816/9d208d9c-773c-4ec5-9469-8cc5aca3b485)

StopMotionCuber:
> > Contributions still welcome? 😬
> 
> In development things go sideways all the time. Better having iterations with mistakes that'd eventually lead to success rather than a flat line ![waveform-path-ecg](https://user-images.githubusercontent.com/4253816/285160155-9d208d9c-773c-4ec5-9469-8cc5aca3b485.svg)

That's true. I'm more comfortable if it goes sideways on my personal projects or work projects that I have control over than on upstream projects.

Anyway, fixing things there and will submit yet another PR when actions are fully running through. Deleted the login of Quay and Dockerhub and pushing to my personal container registry, so I have tested it end to end.
As a sidenode, when thinking I fixed everything, yet another bug appeared with `podman manifest push` needing 2 arguments. Not the case on my personal workstation, as that seemed to be changed with `podman 4.6.0`, which is not available in the Ubuntu 22.04 image.

StopMotionCuber:
Vote for closing this as pipelines seem to be running again

PR #563, opened 2023-11-22
Fix manifest not being created correctly for multi-arch build

StopMotionCuber:
I've seen the latest issue about the manifest not being created after merging #562 and wanted to provide a fix for that. I'm sorry about pipelines that are still failing.

As that action is not running when in a PR, I've created an additional branch on my fork where this change is tested, without pushing the manifest to the repository.
Please wait for [this run](https://github.com/StopMotionCuber/pulp-oci-images/actions/runs/6959530178) to successfully create the manifests in the "Test manifest build" step until merging

PR #566, opened 2023-11-23
Support multi-arch builds

StopMotionCuber:
Trying a fresh run for supporting multi-arch builds, see #546, #562, #563, #564 and #565 for some reference.

In this run, I've tested the changes end to end, with additional modifications (not present in this PR).
[Here](https://github.com/StopMotionCuber/pulp-oci-images/compare/multi-arch...StopMotionCuber:pulp-oci-images:multi-arch-testing) you can take a look at the additional modifications that went into my branch (not using dockerhub and quay.io, only ghcr.io, where the path was modified to my personal namespace).
[The Action for that was passing.](https://github.com/StopMotionCuber/pulp-oci-images/actions/runs/6968129421)

You can see published packages [here](https://github.com/stopmotioncuber?tab=packages). To inspect them, run e.g.
```bash
podman manifest inspect ghcr.io/stopmotioncuber/pulp:3.41.1
```

There are issues with the base image, which is only published for `amd64`. I identified the issue, it's that the according Dockerfile is using
```Dockerfile
FROM docker.io/centos/nginx-116-centos7:1.16
```
as base. This image is not multi-arch (and therefore neither are their descendants) and has seen the last update [two years ago](https://hub.docker.com/r/centos/nginx-116-centos7) according to DockerHub. Is anyone still using the web images? They don't seem _that_ well maintained.

### DIFF
--- a/.github/workflows/pulp_images.yml
+++ b/.github/workflows/pulp_images.yml
@@ -103,8 +103,12 @@ jobs:
         run: |
           podman version
           buildah version
-          podman build --format docker --file images/Containerfile.core.base --tag pulp/base:${TEMP_BASE_TAG} .
-          podman build --format docker --file images/pulp_ci_centos/Containerfile --tag pulp/pulp-ci-centos:${TEMP_BASE_TAG} --build-arg FROM_TAG=${TEMP_BASE_TAG} .
+          sudo podman run --rm --privileged multiarch/qemu-user-static --reset -p yes
+          for ARCH in arm64 amd64
+          do
+            podman build --platform linux/$ARCH --format docker --file images/Containerfile.core.base --tag pulp/base:${TEMP_BASE_TAG}-${ARCH} .
+            podman build --platform linux/$ARCH --format docker --file images/pulp_ci_centos/Containerfile --tag pulp/pulp-ci-centos:${TEMP_BASE_TAG}-${ARCH} --build-arg FROM_TAG=${TEMP_BASE_TAG}-${ARCH} .
+          done
       # we use the docker format (default), even though it may not be the fastest,
       # because it supports saving both images at once.
       # However, it seems to export the common layers twice.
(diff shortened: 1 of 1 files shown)


.github/workflows/pulp_images.yml
```diff
@@ -324,8 +332,8 @@ jobs:
                 tags="${BASE_BRANCH} ${BASE_VERSION}"
               fi
               for tag in $tags; do
-                podman tag pulp/${image_name_looped}:${TEMP_BASE_TAG} ${registry}/pulp/${image_name_looped}:${tag}
-                podman push ${registry}/pulp/${image_name_looped}:${tag}
+                podman manifest create ${registry}/pulp/${image_name_looped}:${tag} containers-storage:localhost/pulp/${image_name_looped}:${TEMP_BASE_TAG}-amd64 containers-storage:localhost/pulp/${image_name_looped}:${TEMP_BASE_TAG}-arm64
+                podman manifest push --all ${registry}/pulp/${image_name_looped}:${tag} ${registry}/stopmotioncuber/${image_name_looped}:${tag}
```
himdel:
```suggestion
                podman manifest push --all ${registry}/pulp/${image_name_looped}:${tag} ${registry}/pulp/${image_name_looped}:${tag}
```

(and on line 370)

StopMotionCuber: Thanks for finding that, fixed it

.github/workflows/pulp_images.yml
```diff
@@ -103,8 +103,12 @@ jobs:
         run: |
           podman version
           buildah version
-          podman build --format docker --file images/Containerfile.core.base --tag pulp/base:${TEMP_BASE_TAG} .
-          podman build --format docker --file images/pulp_ci_centos/Containerfile --tag pulp/pulp-ci-centos:${TEMP_BASE_TAG} --build-arg FROM_TAG=${TEMP_BASE_TAG} .
+          sudo podman run --rm --privileged multiarch/qemu-user-static --reset -p yes
+          for ARCH in arm64 amd64
```
decko, line 107: Can we have different workflows/steps for each arch? So we don't have to use more bash magic here...

StopMotionCuber: I'm not too familiar with GitHub Actions, but we would need to shift the complexity to some kind of fanout and collection (multiple artifacts via `podman save` instead of one including ID management, gathering all the images at one place for manifest push in the end).
I don't know whether and/or how that is possible with a single definition of the desired architectures.

So IMO we would just exchange bash magic with Github Actions magic, which I think wouldn't be worth it

lubosmj (approved):
Looks good to me from the index building and pushing perspective. ARM64 and AMD64 images should be part of one OCI index.

.github/workflows/pulp_images.yml
mikedep333: @StopMotionCuber I agree, let's stick with this for now.

mikedep333 (approved):

StopMotionCuber:
Thanks for merging this 🎉 
CI seems to have failed again after merging, but at least it looks like a failure I'm not responsible for

PR #575, opened 2024-01-14
Bump base image to EL9

dralley:


dralley:
The issue seems to be that something is unhappy with postgresql 13, which is strange because it was already using postgresql 13 before, and all of the migrations complete successfully.

The error happens here: https://github.com/pulp/pulp-oci-images/actions/runs/7517644323/job/20464231737#step:20:1108

Is this at all relevant? https://github.com/pulp/pulp-oci-images/blob/latest/.github/workflows/pulp_images.yml#L244-L250

dralley:
@decko ^

decko:
Gonna ask @mikedep333 about it.

PR #573, opened 2024-01-10
Support gunicorn worker recycling

markafarrell:
Currently the memory usage of the multi-process container increases over time the majority contributor is the gunicorn workers for  pulpcore-api

The suggested fix for these type of issues seems to be to enable worker recycling. 

pulp/pulpcore/issues/2250

The suggested solution of editing `gunicorn.conf.py`, however, does not work (I think due to how s6-overlay works)

This changes adds some additional environment variables that can be use to configure gunicorn worker recycling for both pulpcore-api and pulpcore-content

PR #581, opened 2024-01-23
Revert "Support gunicorn worker recycling"

mdellweg:
Reverts pulp/pulp-oci-images#573


mikedep333:
Sorry @markafarrell , we need to revert this until a new PR is opened up.

The problem is that our CI for plugins uses the current Dockerfiles (pulp_ci_centos) to test installing older versions of pulpcore as well.

The entrypoint changed from `gunicorn` / `pulp-content` to `pulpcore-content` / `pulpcore-api` in 3.35. But the old binaries do not support these arguments, thus the CI breaks.

So we need to only add these arguments if the new binaries are in use.

mdellweg:
I do believe the thing is that these specific options were added even later.

PR #587, opened 2024-02-02
Fix pulp CI tests failing for 2 different reasons

### LINKED ISSUE #584: Pulp docker image tags for releases after pulp 3.43.1 missing
At https://hub.docker.com/r/pulp/pulp docker image tags for pulp releases after 3.43.1 are missing:
https://hub.docker.com/r/pulp/pulp/tags?page=1&name=3.4

### LINKED ISSUE #588: pulp_tests.sh in CI fails to build openshift-examples/web.git
Our CI tests for the Pulp images tries and fails to build https://github.com/openshift-examples/web.git for about the last month or so. We build it because it is an example of running buildah within a container. In other words, nested buildah.

This appears to be due to upstream breakage, building an older commit works.

mikedep333:
Fixes: #584

Fixes: #588

Note: Galaxy tests are still failing due to python version, which is a much larger endeavor under way.

### DIFF
--- a/images/s6_assets/pulp_tests.sh
+++ b/images/s6_assets/pulp_tests.sh
@@ -55,7 +55,12 @@ podman exec -u pulp pulp bash -c "pulpcore-manager add-signing-service --class d
 
 # Test buildah for pulp_container's usage
 podman exec -u pulp pulp podman system migrate
-podman exec -u pulp pulp podman build https://github.com/openshift-examples/web.git
+# This is currently not working due to a bug in it, so workaround below
+# podman exec -u pulp pulp podman build https://github.com/openshift-examples/web.git
+podman exec -u pulp pulp git clone https://github.com/openshift-examples/web.git /tmp/web
+podman exec -u pulp pulp git -C /tmp/web checkout 7a5e95bbf7111f32be27b5fc9bc0070f844a03a3
+podman exec -u pulp pulp podman build /tmp/web
+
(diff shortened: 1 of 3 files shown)


netsandbox:
If you install now pulp-cli with pip, is then the git clone of the pulp-cli repo still needed?

mikedep333:
> If you install now pulp-cli with pip, is then the git clone of the pulp-cli repo still needed?

Hey @netsandbox there's probably a more efficient way to do it. But we still install the test_requirements .

mikedep333:
@netsandbox Also, when I tried installing the folder, it didn't work. It thought it was installing package "UNKNOWN" . And the `pulp` binary was still not found.

images/s6_assets/pulp_tests.sh
```diff
@@ -21,6 +21,7 @@ else
   cd pulp-cli
 fi
 pip install -r test_requirements.txt || pip install --no-build-isolation -r test_requirements.txt
```
mdellweg, line 23: Is this script meant to be run inside the container? Let's not do that.
By calling pip, we may end up with an incompatible version of pulp-glue for pulpcore.

Instead, I suggest: Install the cli, retrieve it's version and checkout the corresponding release tag to grant access to the matching test suite.

mikedep333: @mdellweg No it's not. It is run on the host.

dkliban (approved):

PR #589, opened 2024-02-06
Drop createrepo_c build deps

dralley:
We use the prebuilt wheels nowadays

[noissue]

### DIFF
--- a/images/Containerfile.core.base
+++ b/images/Containerfile.core.base
@@ -36,12 +36,6 @@ RUN dnf -y module enable nginx:1.22
 
 # glibc-langpack-en is needed to provide the en_US.UTF-8 locale, which Pulp
 # seems to need.
-#
-# The last 5 lines (before clean) are needed until python3-createrepo_c gets an
-# RPM upgrade to 0.16.2. Until then, we install & build it from PyPI.
-#
-# TODO: Investigate differences between `dnf builddep createrepo_c` vs the list
-# of dependencies below. For example, drpm-devel.
 RUN dnf -y install python${PYTHON_VERSION} python3-cryptography python${PYTHON_VERSION}-devel && \
     dnf -y install openssl openssl-devel && \
     dnf -y install openldap-devel --exclude selinux* && \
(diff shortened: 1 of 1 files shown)


dralley:
Oh, right, we can't do this until we upgrade to Python 3.9, because createrepo_c doesn't publish wheels for 3.8

images/Containerfile.core.base
```diff
@@ -45,15 +39,7 @@ RUN dnf -y install python3 python3-cryptography python3-devel && \
     dnf -y install python3-libcomps && \
     dnf -y install libpq-devel && \
```
dralley, line 48: Do we remember what this is required for?  And openldap-devel?

dralley:
@mikedep333 I think this is failing because we're building aarch64 images now, but the binary packages don't exist for createrepo_c, solv, and maybe libcomps, so it falls back to trying to build from source.

I guess the only alternatives are qemu, or waiting until GHA has ARM runners available.

images/Containerfile.core.base
dralley: @mikedep333 @mdellweg

decko:
> @mikedep333 I think this is failing because we're building aarch64 images now, but the binary packages don't exist for createrepo_c, solv, and maybe libcomps, so it falls back to trying to build from source.
> 
> I guess the only alternatives are qemu, or waiting until GHA has ARM runners available.

And look what we got https://github.blog/2024-06-03-arm64-on-github-actions-powering-faster-more-efficient-build-systems/

dralley:
But not on free plans yet.  And older branches use older versions of createrepo_c and libcomps and such, for which no binary arm packages are uploaded, so we might have to wait a bit for some of those to phase out.

images/Containerfile.core.base
```diff
@@ -53,15 +47,7 @@ RUN dnf -y install python${PYTHON_VERSION} python3-cryptography python${PYTHON_V
...
-    dnf -y install swig && \
-    dnf -y install podman fuse-overlayfs buildah --exclude container-selinux && \
-    dnf -y install xz && \
-    dnf -y install libmodulemd-devel && \
-    dnf -y install libcomps-devel && \
-    dnf -y install zchunk-devel && \
-    dnf -y install ninja-build && \
-    dnf -y install cairo-devel cmake gobject-introspection-devel cairo-gobject-devel && \
-    dnf -y install libcurl-devel libxml2-devel sqlite-devel file-devel && \
+-   dnf -y install podman fuse-overlayfs buildah --exclude container-selinux && \
```
gerrod3: You have a dash here on line 50 that needs to be removed.

gerrod3 (approved):
Nice!

PR #591, opened 2024-02-08
Add lsof to the base image installs

gerrod3:
This line in the plugin template (https://github.com/pulp/plugin_template/blob/main/templates/github/.github/workflows/scripts/before_script.sh.j2#L37) slows the down the CI by 30sec ~ 1min because dnf has to update its indexes before installing. We already have `which` so adding `lsof` will allow us to remove those lines in the plugin_template.

### DIFF
--- a/images/Containerfile.core.base
+++ b/images/Containerfile.core.base
@@ -62,6 +62,7 @@ RUN dnf -y install python3 python3-cryptography python3-devel && \
     dnf -y install jq && \
     dnf -y install which && \
     dnf -y install rpm-sign && \
+    dnf -y install lsof && \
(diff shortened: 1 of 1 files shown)


dralley:
@gerrod3 You will need to rebase

mikedep333 (approved):

mikedep333:
@gerrod3 If it's only needed by CI tests, we can also just add it to pulp_ci_centos instead of base.

gerrod3:
> @gerrod3 If it's only needed by CI tests, we can also just add it to pulp_ci_centos instead of base.

We already have all the other external test dependencies in the base image, what is one more?

PR #592, opened 2024-02-08
Bump base images to use EL9 

### LINKED ISSUE #526: Regex issue in postgres-init script
**Version**
Using the latest code and to build my own images via the Dockerfiles

**Describe the bug**
In the `images/s6_assets/init/postgres-init` script theres a line `postgres --version | egrep -o "1[0-9]"` which returns 13 and 10 which then causes the script to attempt to upgrade postgres which then fails.

This is because:
postgres --version returns:
`postgres (PostgreSQL) 13.10` and the regex matches 2 numbers.

**To Reproduce**
```
docker build -t pulp/base -f images/Containerfile.core.base .
docker build -t pulp/pulp-ci-centos -f images/pulp_ci_centos/Containerfile .
docker build -t pulp/stable -f images/pulp/stable/Containerfile .
```

**Expected behavior**
When I run the pulp/stable container I want it to start up without any database issues.

**Additional context**
Add any other context about the problem here.  Please provide links to any previous discussions via Discourse or Bugzilla.

### LINKED ISSUE #537: Bump base image to use EL9
After some discussion about possible updates to our EL8 based image([here](https://github.com/pulp/pulp-oci-images/issues/506) and [here](https://matrix.to/#/!aVApiNMtnstWbwDcVU:matrix.org/$PSV45LKnuFIsr2u151-NZYWaxwvjYPRso4jpYc6iHW0?via=libera.chat&via=matrix.org&via=fedora.im)) seems this could be a good time to use EL9 as the base to our container images and Python 3.9 as base Python version for Pulp.

This could improve our images with recent versions for most of our dependencies, better tooling and so on.

mikedep333:
closes #537

Also: Fix a regex issue that impedes swapping OS version
closes #526

### DIFF
--- a/images/s6_assets/init/postgres-init
+++ b/images/s6_assets/init/postgres-init
@@ -13,7 +13,7 @@ export C035="\e[35m"
 export C036="\e[36m"
 export C037="\e[37m"
 
-PGVERSION=$(postgres --version | egrep -o "1[0-9]")
+PGVERSION=$(postgres --version | egrep -o "1[0-9]" | head -n1)
 PGHOME="/var/lib/pgsql"
 PGDATA="${PGHOME}/data"
 PGDATA_OLD="${PGHOME}/data_old"
--- a/.github/workflows/pulp_images.yml
+++ b/.github/workflows/pulp_images.yml
@@ -103,7 +103,7 @@ jobs:
           for ARCH in arm64 amd64
           do
             podman build --platform linux/$ARCH --format docker --file images/Containerfile.core.base --tag pulp/base:${TEMP_BASE_TAG}-${ARCH} .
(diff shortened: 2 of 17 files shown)


images/s6_assets/init/postgres-init
```diff
@@ -13,7 +13,7 @@ export C035="\e[35m"
 export C036="\e[36m"
 export C037="\e[37m"
 
-PGVERSION=$(postgres --version | grep -E -m 1 -o "1[0-9]")
```
dralley: the `-m 1` originally came from here: https://github.com/pulp/pulp-oci-images/pull/522/files

If it doesn't work, then you can probably drop it completely

dralley:
I can't remember why I added https://github.com/pulp/pulp-oci-images/pull/592/commits/1783c61a56a59668ee4907bfe4c26fca9a920491

Fix regex check can probably be merged into the second commit

gerrod3 (approved):

PR #597, opened 2024-02-16
Update CI branches

gerrod3:
@mikedep333 We have not been creating the release branches for the last ~20 releases. We need to create branches for 3.39 and 3.45. What commits should we use for these branches? I think we should choose commits around their release dates: 3.39 - 10/25/23 , 3.45 - 01/24/24.

### DIFF
--- a/.ci/scripts/update_ci_branches.py
+++ b/.ci/scripts/update_ci_branches.py
@@ -3,8 +3,6 @@ import requests
 from yaml import safe_load
 from packaging import version
 
-branches = ["3.28", "3.23", "3.22", "3.21", "3.18", "3.16"]
-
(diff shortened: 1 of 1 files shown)


.ci/scripts/update_ci_branches.py
```diff
@@ -1,9 +1,8 @@
 import os
 import requests
-from yaml import safe_load
 from packaging import version
 
-branches = ["3.28", "3.23", "3.22", "3.21", "3.18", "3.16"]
+branches = ["3.45", "3.39", "3.28", "3.23", "3.22", "3.21", "3.18"]
```
mdellweg: Maybe we should take these from pulpcore's template plugin.

mikedep333 (approved):

PR #602, opened 2024-02-21
Stop building createrepo_c

mdellweg:
[noissue]


dralley:
@mdellweg You've duplicated https://github.com/pulp/pulp-oci-images/pull/589 :P

However this is different because you're installing createrepo_c from RHEL 9.  That won't work, their version is still too old for us, it doesn't have zstd support for example.

What we'd need to do is backport zstd support (and maybe some other things) to a pre-1.0 release that they can ship in RHEL 9... and then wait like 6 months.

mdellweg:
Hah. It was worth a try for about half a GB of container size. Closing this one.

PR #604, opened 2024-02-28
Get correct version of pulp-cli tests

### LINKED ISSUE #603: OCI images cli tests keep failing.
Instead, I suggest: Install the cli, retrieve it's version and checkout the corresponding release tag to grant access to the matching test suite.

_Originally posted by @mdellweg in https://github.com/pulp/pulp-oci-images/pull/587#discussion_r1477035103_

Currently we run the wrong test suite against the installed version of the CLI, incidentally failing for that reason.

markafarrell:
Fixes #603

### DIFF
--- a/images/s6_assets/pulp_tests.sh
+++ b/images/s6_assets/pulp_tests.sh
@@ -12,16 +12,26 @@ echo "Base Address: $BASE_ADDR"
 grep "127.0.0.1   pulp" /etc/hosts || echo "127.0.0.1   pulp" | sudo tee -a /etc/hosts
 
 echo "Installing Pulp-CLI"
+pip install pulp-cli
+
+# Retreive installed pulp-cli version
+PULP_CLI_VERSION=$(python3 -c \
+  'import importlib.metadata; \
+   from packaging.version import Version; \
+   print(Version(importlib.metadata.version("pulp-cli")))')
+
+# Checkout git repo for pulp-cli at correct version to fetch tests
 if [ -d pulp-cli ]; then
   cd pulp-cli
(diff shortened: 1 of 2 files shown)


markafarrell:
We have an issue as CI is attempting to use `pulp/pulp-web:3.48.0` but this does not exist on dockerhub.

Not sure how we resolve this.

markafarrell:
> We have an issue as CI is attempting to use `pulp/pulp-web:3.48.0` but this does not exist on dockerhub.
> 
> Not sure how we resolve this.

Seems to have resolved itself. 

😃

images/s6_assets/pulp_tests.sh
```diff
@@ -12,16 +12,25 @@ echo "Base Address: $BASE_ADDR"
...
+
+# Retreive installed pulp-cli version
+PULP_CLI_VERSION=$(python3 -c \
+  'import importlib.metadata; \
+   from packaging.version import Version; \
+   print(Version(importlib.metadata.version("pulp-cli")))')
+
+# Checkout git repo for pulp-cli at correct version to fetch tests
+if [ ! -d pulp-cli ]; then
   git clone --depth=1 https://github.com/pulp/pulp-cli.git
```
mdellweg:
```suggestion
git clone --depth=1 https://github.com/pulp/pulp-cli.git -b "${PULP_CLI_VERSION}"
```
You should be able to clone that tag directly.

markafarrell: Thanks. I have made the change

mdellweg (approved):

PR #606, opened 2024-02-29
Add ability to configure python version in base image

gerrod3:
Big refactor in the CI. I've broken up `pulp_images.yml` into multiple components:

- The CI tests for pull-requests -> `ci.yml`
- The release machinary -> `release.yml`

The common actions between each run (building images and publishing) have been converted to actions
- Building the base images -> `actions/base_images`
- Building an app image -> `actions/build_image`
- Publishing to the registries -> `actions/publish_images`

There is still a big overlap in `ci.yml` & `release.yml` that I want to remove, but I've been writing this for a while so I need to get some feedback before trying to make it perfect.


gerrod3:
Got the tests to pass, but I am not sure if the release machinary will work. Need to double-check to make sure the tagging logic is correct.

.github/actions/base_images/action.yml
```diff
@@ -0,0 +1,62 @@
+---
+name: Build Base Images
```
mikedep333, line 2: Nice! I'm very happy we have these separated out into actions.

mikedep333 (approved):
Thank you so much for all of this! I remember the comments from the meeting recently where you explained the design decisions (like creating a test script).

.github/actions/test_image/action.yml
```diff
@@ -0,0 +1,55 @@
...
+    default: 'stable'
+  app_branch:
+    description: 'The branch the app was built on'
+    required: true
+  app_arch_tag:
+    description: 'The temporary amd64 arch tag of the built app image'
+    required: true
+
+runs:
+  using: "composite"
```
mikedep333, line 20: Awesome! release.yml and ci.yml are pretty short now.

mikedep333 (approved):

PR #623, opened 2024-04-03
Configure NGINX using API_ROOT & CONTENT_PATH_PREFIX

### LINKED ISSUE #605: Configure NGINX using values of API_ROOT and CONTENT_PREFIX
When the all in one container starts up, it should check what the values of API_ROOT and CONTENT_PREFIX are. It should adjust the nginx config to match those values and then restart nginx.

gerrod3:
fixes: #605


images/s6_assets/template_nginx.py
```diff
@@ -0,0 +1,37 @@
...
+    parser = argparse.ArgumentParser(
+        description="Create Pulp's nginx conf file based on current settings.",
+    )
+    parser.add_argument("template_file", type=open)
+    parser.add_argument("output_file", type=argparse.FileType("w"))
+    args = parser.parse_args()
+
+    https = os.getenv("PULP_HTTPS", "false")
+    values = {
+        "https": https.lower() in ("true", "t", "yes", "y", "1"),
```
mdellweg: Are these values documented anywhere?
Do _all_ the other places recognize them in the same way?

gerrod3: No, previously we were only checking for 'true' after lower-casing.

dkliban (approved):

dkliban (changes requested):
Let's just check for value of 'true' for now. i'll add a new issue to allow all our boolean settings to accept a broader set of values.

dkliban (approved):
Awesome! Thanks for this great feature! 

I filed https://github.com/pulp/pulp-oci-images/issues/624 to track the improvement for the boolean settings.

mdellweg (approved):
This is a nice solution.

However (and this PR is not alone), it is kind of weird how we stretch the container idea. This almost looks like self modifying code.

Issue #634, opened 2024-06-04
Old CI CentOS 8 images fail to use repos


mikedep333:
CentOS Stream 8 is end of life, and we have to update the container images on branches 3.39, 3.28, 3.22, and 3.21 to use vault.centos.org as the repo.

PR #638, opened 2024-06-04
3.21: Update CentOS 8 containers for the distro's EOL

mikedep333:
Fixes: #634

### DIFF
--- a/CHANGES/584.bugfix
+++ b/CHANGES/584.bugfix
@@ -1,0 +1,1 @@
+Fix pulp CI tests failing with "pulp: command not found"
--- a/CHANGES/634.bugfix
+++ b/CHANGES/634.bugfix
@@ -1,0 +1,1 @@
+Update CentOS 8 containers for the distro's EOL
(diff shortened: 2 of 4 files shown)


decko (dismissed):

dralley (approved):

dralley:
would it be feasible to switch to alma instead of using the vault repos?

mikedep333:
> would it be feasible to switch to alma instead of using the vault repos?

It would probably be more work in the short-term, but help us long-term.

mikedep333:
@decko The tests are not passing anyway on nightly CI. How about we just merge this or close this?

lubosmj (approved):

lubosmj:
If the release image runner is failing, how can we ensure that this change will be propagated to the upstream? I wonder if we should at least temporarily merge this instead: https://github.com/pulp/plugin_template/pull/873.

decko:
> If the release image runner is failing, how can we ensure that this change will be propagated to the upstream? I wonder if we should at least temporarily merge this instead: [pulp/plugin_template#873](https://github.com/pulp/plugin_template/pull/873).

What do you think @mikedep333? 
Maybe some coordination on the #pulp-deployers channel to understand why some checks are failing and which PR should be prioritized.

mikedep333:
@lubosmj @decko The 3.21 branch has been failing to publish for the last 6 months. Are you sure you guys actually use it?

https://quay.io/repository/pulp/pulp-ci-centos?tab=tags

lubosmj:
We are using the `latest` tag. Which was uploaded 4 months ago:
image

PR #643, opened 2024-06-23
Added missing 'pinentry' package

COM8:
This package is required to import GPG keys via e.g. `gpg --import /var/lib/pulp/gpg/ppos_rpm_private.gpg`  inside the image.

Else it will fail with the following error:
```
pulp bash -c 'gpg --import /var/lib/pulp/gpg/ppos_rpm_private.gpg'
gpg: key D6763692AE005CBF: "..." not changed
gpg: key D6763692AE005CBF/D6763692AE005CBF: error sending to agent: No pinentry
gpg: error reading '/var/lib/pulp/gpg/ppos_rpm_private.gpg': No pinentry
gpg: import from '/var/lib/pulp/gpg/ppos_rpm_private.gpg' failed: No pinentry
gpg: Total number processed: 0
gpg:              unchanged: 1
gpg:       secret keys read: 1
```

### DIFF
--- a/images/Containerfile.core.base
+++ b/images/Containerfile.core.base
@@ -71,6 +71,7 @@ RUN dnf -y install python${PYTHON_VERSION} python3-cryptography python${PYTHON_V
     dnf -y install which && \
     dnf -y install rpm-sign && \
     dnf -y install lsof && \
+    dnf -y install pinentry && \
     getcap /usr/bin/newuidmap  | grep cap_setuid || dnf -y reinstall -y shadow-utils && \
     dnf clean all
 


mdellweg:
I don't think this is a good addition to the container. If you need it for that command, you can install it there. But i suspect, a key with a password will not work nicely for any sort of automated package/release signing in pulp.

COM8:
I see your point there. But in my eyes an officially maintained package with a size of [`256.13 KB`](https://fedora.pkgs.org/40/fedora-x86_64/pinentry-1.3.0-2.fc40.x86_64.rpm.html) and no real dependencies that helps handling encrypted keys is a good addition.

As far as I understand it, after importing the password encrypted GPG key to my pulp container I can remove/cache the password for signing.

PR #648, opened 2024-07-09
Fix docker-compose add_signing_service script

### LINKED ISSUE #647: Unable to launch docker compose stack
**Version**
Latest version from Git repository

**Describe the bug**
Starting the docker compose stack as per [documentation](https://pulpproject.org/pulp-oci-images/docs/admin/tutorials/quickstart/#running-with-docker-and-scaling)  with the command `docker compose up` results in a crash with code:
```
set_init_password_service-1 exited with code 0
Gracefully stopping... (press Ctrl+C again to force)
service "signing_key_service" didn't complete successfully: exit 1
```

checking the logs for the container `signing_key_service`
```
$ docker logs 476042309d92
Waiting on postgresql to start...
Postgres started.
Checking for database migrations
Database migrated!
/usr/bin/add_signing_service.sh: line 6: PULP_SIGNING_KEY_FINGERPRINT: unbound variable
Waiting on postgresql to start...
Postgres started.
Checking for database migrations
Database migrated!
/usr/bin/add_signing_service.sh: line 6: PULP_SIGNING_KEY_FINGERPRINT: unbound variable
```
----

Running the container stack again with `docker compose up -d` to start the stack in the background results service `signing_key_service` not being able to start

```
[+] Running 9/9
 ✔ Container compose-redis-1                      Healthy                                                                                    7.5s
 ✔ Container compose-postgres-1                   Healthy                                                                                    7.5s
 ✔ Container compose-set_init_password_service-1  Exited                                                                                     9.5s
 ✔ Container compose-migration_service-1          Exited                                                                                     7.5s
 ✔ Container compose-pulp_content-1               Running                                                                                    0.0s
 ✔ Container compose-pulp_content-2               Running                                                                                    0.0s
 ✘ Container compose-signing_key_service-1        service "signing_key_service" didn't complete...                                          12.6s
 ✔ Container compose-pulp_worker-1                Running                                                                                    0.0s
 ✔ Container compose-pulp_worker-2                Running                                                                                    0.0s
service "signing_key_service" didn't complete successfully: exit 1
```

**To Reproduce**
Follow the documentation (note that the docs uses outdated `docker-compose` which is not supported)

```
git clone git@github.com:pulp/pulp-oci-images.git
cd images/compose
docker-compose up
```

**Expected behavior**
I expect the stack to be running after running `docker compose up` and NOT crash / stop unexpectedly.

gerrod3:
fixes: #647

### DIFF
--- a/images/assets/add_signing_service.sh
+++ b/images/assets/add_signing_service.sh
@@ -3,7 +3,7 @@
 /usr/bin/wait_on_postgres.py
 /usr/bin/wait_on_database_migrations.sh
 
-if [ -n "${PULP_SIGNING_KEY_FINGERPRINT}" ]; then
+if [ -n "${PULP_SIGNING_KEY_FINGERPRINT-}" ]; then
   /usr/local/bin/pulpcore-manager add-signing-service "${COLLECTION_SIGNING_SERVICE}" /var/lib/pulp/scripts/collection_sign.sh "${PULP_SIGNING_KEY_FINGERPRINT}"
   /usr/local/bin/pulpcore-manager add-signing-service "${CONTAINER_SIGNING_SERVICE}" /var/lib/pulp/scripts/container_sign.sh "${PULP_SIGNING_KEY_FINGERPRINT}" --class container:ManifestSigningService
 fi
--- a/CHANGES/647.bugfix
+++ b/CHANGES/647.bugfix
@@ -1,0 +1,1 @@
+Fixed default docker compose file not launching when PULP_SIGNING_KEY_FINGERPRINT was unset.


images/assets/add_signing_service.sh
```diff
@@ -3,7 +3,7 @@
 /usr/bin/wait_on_postgres.py
 /usr/bin/wait_on_database_migrations.sh
 
-if [ -n "${PULP_SIGNING_KEY_FINGERPRINT}" ]; then
+if [ -n "${PULP_SIGNING_KEY_FINGERPRINT-}" ]; then
```
mdellweg, line 6: Isn't that:
```suggestion
if [ -n "${PULP_SIGNING_KEY_FINGERPRINT:-}" ]; then
```

gerrod3: Adding the colon makes it also a nullity check. Do you think we need to check also if the value is null, since the default I am setting it to is null? (I am actually not sure how null works in bash land)

https://devhints.io/bash#parameter-expansions

mdellweg: Well, I'd say if the variable was not set or set to "" it is safe to assume we cannot configure a signing service from it.
But honestly i didn't know about "-" alone.
But now reading up, i think both produce the same result here, as we replace with "".

mdellweg (approved):

PR #657, opened 2024-08-20
Remove deprecated "update-ca-trust force-enable" call

lubosmj:
According to the docs, it is sufficient just to call "update-ca-trust [extract]" which instructs update-ca-trust to scan the SOURCE CONFIGURATION and produce updated versions of the consolidated configuration files stored below the /etc/pki/ca-trust/extracted directory hierarchy.

https://www.unix.com/man-page/centos/8/update-ca-trust/

[noissue]

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -113,6 +113,8 @@ jobs:
           podman ps -a
           podman images -a
           podman logs pulp
+          echo "plottin'"
+          cat /var/lib/pgsql/data/log/*.log
           cd images/compose
           podman-compose logs
           podman logs --tail=10000 compose_pulp_api_1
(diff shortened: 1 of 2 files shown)


ggainey:
And also, `force-enable` **no longer exists** as a subcommand.

ggainey (dismissed):

git-hyagi (dismissed):

ggainey:
Maybe "Remove" the old image reference? :)

lubosmj:
> Maybe "Remove" the old image reference? :)

Do you mean this?
image

If yes, it did not help.

ggainey:
> Do you mean this?

I just meant the commit-msg - instead of "Remote old image".  But if it isn't the problem, then that's not, alas, relevant.

PR #661, opened 2024-09-11
Fix the failing CI

lubosmj:


images/pulp_ci_centos/Containerfile
```diff
@@ -2,6 +2,8 @@ ARG FROM_TAG="latest"
 FROM pulp/base:${FROM_TAG}
 # https://ryandaniels.ca/blog/docker-dockerfile-arg-from-arg-trouble/
 
+RUN dnf -y install xz
```
mdellweg, line 5: It seems to be good practice not to leave any cache files around in _any_ layer of the image.

```suggestion
RUN dnf -y install xz && dnf clean all
```

Maybe it's even worth combining this with the dnf command ion line 25.

lubosmj: Yeah, but still, the root cause of the problem has not been resolved. Neither writing to `/usr/share/containers/storage.conf` nor `$HOME/.config/containers/storage.conf` helped. It looks like the `storage.conf` file ceased to exist.

PR #663, opened 2024-09-17
[Backport 3.49] Fix base image build having a selinux dependency collision

gerrod3:
This is to check if the CI issue in #662 is due to the fact that the CI has been broken for 2 weeks and we don't have the latest images for it to pull from.

### DIFF
--- a/images/Containerfile.core.base
+++ b/images/Containerfile.core.base
@@ -44,7 +44,7 @@ RUN dnf -y module enable nginx:1.22
 # of dependencies below. For example, drpm-devel.
 RUN dnf -y install python${PYTHON_VERSION} python3-cryptography python${PYTHON_VERSION}-devel && \
     dnf -y install openssl openssl-devel && \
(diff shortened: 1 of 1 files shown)


images/Containerfile.core.base
```diff
@@ -54,7 +54,7 @@ RUN dnf -y install python${PYTHON_VERSION} python3-cryptography python${PYTHON_V
     dnf -y install libpq-devel && \
     dnf -y install python3-setuptools && \
     dnf -y install swig && \
-    dnf -y install buildah --exclude container-selinux && \
+    dnf -y install podman fuse-overlayfs buildah --exclude container-selinux && \
```
mdellweg, line 57: Sounds good to me.
Is this selinux stuff a weak dependency? Should we think about https://discussion.fedoraproject.org/t/which-kind-of-dependencies-suggested-recommended-does-dnf-install/74111 in general?

mdellweg (approved):

PR #670, opened 2024-09-26
Make CI smarter about when to rebuild the base images

gerrod3:
The base images action now takes the hash of the files used to build the base-image & ci-centos-image and then uses that as the key to pull the images from the cache. If it is a cache miss, then we build the images and upload to the cache like normal. If it is a cache hit then we import the images and check if each one needs to be rebuilt by running `dnf check-update`. This should reduce the amount of times we build the base & ci-centos images which is roughly half the CI run time! Also, I removed the temporary build tags as they were superfluous (and confusing).

### DIFF
--- a/.github/actions/base_images/action.yml
+++ b/.github/actions/base_images/action.yml
@@ -1,15 +1,15 @@
 ---
 name: Build Base Images
-description: Build the base images (pulp/base & pulp/pulp-ci-centos9)
+description: Build the base images (pulp/base & pulp/pulp-ci-centos9) if needed
 # Both ARM64 & x86-64 versions of each are built
-# Save the images to a tar and upload to a cache using output 'pulp_ci_centos_id' as the key
+# Use hashFiles(base_image_files, pulp-ci_image_files) + python_version as the key to the cache
 inputs:
   python_version:
     required: true
     description: "Python Version to use to build, e.g '3.9'"
 outputs:
-  pulp_ci_centos_id:
-    value: ${{ steps.pulp_ci_centos_id.outputs.pulp_ci_centos_id }}
+  base_cache_key:
+    value: ${{ steps.hash_key.outputs.base_cache_key }}
     description: "The cache key the built images were uploaded to."
 
 runs:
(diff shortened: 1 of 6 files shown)


gerrod3:
Current base-images CI results: first build (cache-miss) -> 18 mins, second build (cache-hit) -> 6 mins. The new slow part in the cache-hit is that each container we are checking have to update their repo metadata which is quite slow... Wonder if there is a way to do this once and reuse the same metadata across containers.

.github/actions/base_images/action.yml
```diff
@@ -1,40 +1,91 @@
...
+      if: steps.cache.outputs.cache-hit == 'true'
+      run: |
+        echo "Base Images were in cache"
+        podman load -i base-images.tar.gz
+      shell: bash
+
+    - name: Check for updates on cached images
+      if: steps.cache.outputs.cache-hit == 'true'
+      run: |
+        sudo podman run --rm --privileged multiarch/qemu-user-static --reset -p yes
```
mdellweg, line 46: What is this for?

.github/actions/base_images/action.yml
```diff
@@ -1,40 +1,91 @@
...
+        images=()
+        for IMAGE in base pulp-ci-centos9; do
+          for ARCH in arm64 amd64; do
+            echo "Checking if rebuild needed for ${IMAGE}-${ARCH}"
+            podman run --pull=never pulp/${IMAGE}:ci-${ARCH} bash -c "dnf check-upgrade"
+            if [ $? -gt 0 ]; then
+              echo "Rebuild needed"
```
mdellweg:
```suggestion
              echo "Rebuild needed for ${IMAGE}-${ARCH}"
```

.github/actions/base_images/action.yml
```diff
@@ -1,40 +1,91 @@
...
         podman version
         buildah version
-        sudo podman run --rm --privileged multiarch/qemu-user-static --reset -p yes
-        for ARCH in arm64 amd64
-        do
-          podman build --platform linux/$ARCH --format docker --file images/Containerfile.core.base --tag pulp/base:${TEMP_BASE_TAG}-${ARCH} --build-arg PYTHON_VERSION=${PYTHON_VERSION} .
-          podman build --platform linux/$ARCH --format docker --file images/pulp_ci_centos/Containerfile --tag pulp/pulp-ci-centos9:${TEMP_BASE_TAG}-${ARCH} --build-arg FROM_TAG=${TEMP_BASE_TAG}-${ARCH} .
+        for image in "${images[@]}"; do
+          echo "Building image ${image}"
+          arch=${image:(-5):5}
```
mdellweg: Dangerous assumption. Not all archs out there have 5 chars.

.github/actions/base_images/action.yml
gerrod3: This line allows us to run/build the arm64 images. Without it podman won't be able to run them.

.github/actions/base_images/action.yml
gerrod3: We are currently only using archs with 5 chars. What would you suggest I change it to?

.github/actions/base_images/action.yml
mdellweg: That at the very least should be a comment here.

Just more background info for me: This runs a command in a container image named "multiarch/qemu-user-static" that flips a switch somewhere and exits right away?

.github/actions/base_images/action.yml
mdellweg: Sooo, image can contain dashes. Maybe we can assume arch does not and split on the last "-". Or we use ":" instead as a separator.

Also I'd really prefer to consistently use capitalized (all caps) variable names in shell scripts.

.github/actions/base_images/action.yml
```diff
@@ -1,40 +1,91 @@
...
+          case $image in
+            base-arm64 | base-amd64)
```
mdellweg:
```suggestion
            base-*)
```

.github/actions/base_images/action.yml
```diff
@@ -1,40 +1,91 @@
...
+              podman build --platform linux/$arch --format docker --file images/Containerfile.core.base --tag pulp/base:ci-$arch --build-arg PYTHON_VERSION=${{ inputs.python_version }} .
```
mdellweg: I know this may seem overly cautious, but shellcheck would suggest it too, for good reasons.
```suggestion
              podman build --platform "linux/${arch}" --format docker --file images/Containerfile.core.base --tag "pulp/base:ci-${arch}" --build-arg PYTHON_VERSION=${{ inputs.python_version }} .
```

.github/actions/base_images/action.yml
gerrod3: Yeah it runs a script in the container that modifies a file on the host to enable support. https://github.com/multiarch/qemu-user-static

.github/actions/base_images/action.yml
gerrod3: I switched the separator to use ":" so now we can add support for archs with more than 5 chars.

mdellweg (approved):
I'm ready to try this.

PR #671, opened 2024-10-01
Fix base-image CI exiting early when update is needed

gerrod3:


.github/actions/base_images/action.yml
```diff
@@ -45,6 +45,8 @@ runs:
         # Enable running/building ARM64 images: https://github.com/multiarch/qemu-user-static
         sudo podman run --rm --privileged multiarch/qemu-user-static --reset -p yes
         IMAGES=()
+        # Github runs with -e by default, but dnf check-upgrade returns 100 when upgrade needed
+        set +e
```
mdellweg: Shouldn't we do
```
if ! podman run --pull=never pulp/${IMAGE}:ci-${ARCH} bash -c "dnf check-upgrade"
then
...
fi
```
instead?

mdellweg (approved):
Looks good now.

gerrod3:
The compose test was failing because it was trying to pull the tag 3.63 for the web image, but we haven't published that tag yet since our CI was broken yesterday during the release. I changed the code so that it'll pull 'latest' instead of the branch version for PRs and releases on latest so that we should be able to pull something. Now how useful it is to test compose with the multi-process images, I don't know.

mdellweg (approved):

PR #672, opened 2024-10-08
Fix 3.21 CI

gerrod3:
Time to finally make these old CIs green


dralley (approved):

gerrod3:
This was a lot more work then I thought it was going to be
image

Finally got CI green. The release workflow could still be wrong, but we won't know till we merge.

dralley (approved):

PR #674, opened 2024-10-09
Fix release failing when there were duplicate tags

gerrod3:
We can get duplicate tags right after we do a release and not immediately merge the version bump in pulpcore. Nightly will then have the same y-version as the latest release which creates duplicate tags in our tagging logic. Now we de-dup the tags before publishing.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -130,6 +130,7 @@ jobs:
               tags="${tags} ${NIGHTLY_VERSION} ${NIGHTLY_BRANCH} latest"
             fi
           fi
+          tags=$(echo "${tags}" | xargs -n1 | sort -u | xargs)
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -130,6 +130,7 @@ jobs:
               tags="${tags} ${NIGHTLY_VERSION} ${NIGHTLY_BRANCH} latest"
             fi
           fi
+          tags=$(echo "${tags}" | xargs -n1 | sort -u | xargs)
```
mdellweg, line 133: I did not know, xargs can be used that way.

mdellweg: according to `man xargs`:
```
[...] and executes the command (default is echo) [...]
```

mdellweg (approved):

PR #677, opened 2024-10-11
Add pulp-ui to pulp image

### LINKED ISSUE #664: Add Pulp UI to All-In-One image
We're working on a UI!! [0][1] We want this UI to present by default in our images, so first step is to add it to the single container image. 

[0] https://discourse.pulpproject.org/t/pulp-ui-working-group-forming/1704/
[1] https://github.com/pulp/pulp-ui

gerrod3:
There's many different ways to serve these files, here's one way I chose. Would like some feedback on this approach.

TODO:

- [x] Add logic in CI to find ui url
- [ ] Add support for nightly image??
- [x] Add support for templating a config file for values like API_ROOT (maybe in a future PR once we decide how we will do this in the ui)
- [x] Add issue changelog

### DIFF
--- a/images/pulp/stable/Containerfile
+++ b/images/pulp/stable/Containerfile
@@ -12,7 +12,9 @@ ARG PULP_MAVEN_VERSION=""
 ARG PULP_PYTHON_VERSION=""
 ARG PULP_RPM_VERSION=""
 ARG PULP_OSTREE_VERSION=""
+ARG PULP_UI_URL=""
 
+ENV PULP_UI=${PULP_UI_URL:-false}
 COPY images/assets/requirements.extra.txt /requirements.extra.txt
 
 RUN pip3 install --upgrade \
@@ -39,3 +41,10 @@ USER root:root
 RUN ln $(pip3 show pulp_ansible | sed -n -e 's/Location: //p')/pulp_ansible/app/webserver_snippets/nginx.conf /etc/nginx/pulp/pulp_ansible.conf
 RUN ln $(pip3 show pulp_container | sed -n -e 's/Location: //p')/pulp_container/app/webserver_snippets/nginx.conf /etc/nginx/pulp/pulp_container.conf
 RUN ln $(pip3 show pulp_python | sed -n -e 's/Location: //p')/pulp_python/app/webserver_snippets/nginx.conf /etc/nginx/pulp/pulp_python.conf
+
+RUN \
+  if [ -n "$PULP_UI_URL" ]; then \
+    mkdir -p "${PULP_STATIC_ROOT}pulp_ui"; \
+    curl -Ls $PULP_UI_URL | tar -xzv -C "${PULP_STATIC_ROOT}pulp_ui"; \
+  fi
+
--- a/images/s6_assets/nginx.conf.j2
+++ b/images/s6_assets/nginx.conf.j2
@@ -105,6 +105,20 @@ http {
 
         include pulp/*.conf;
 
+        {% if ui | default(false) -%}
+        location /static/pulp_ui/ {
+            root {{ pulp_ui_static|replace("static/pulp_ui/", "") }};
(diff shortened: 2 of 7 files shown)


images/pulp/stable/Containerfile
```diff
@@ -39,3 +41,10 @@ USER root:root
 RUN ln $(pip3 show pulp_ansible | sed -n -e 's/Location: //p')/pulp_ansible/app/webserver_snippets/nginx.conf /etc/nginx/pulp/pulp_ansible.conf
 RUN ln $(pip3 show pulp_container | sed -n -e 's/Location: //p')/pulp_container/app/webserver_snippets/nginx.conf /etc/nginx/pulp/pulp_container.conf
 RUN ln $(pip3 show pulp_python | sed -n -e 's/Location: //p')/pulp_python/app/webserver_snippets/nginx.conf /etc/nginx/pulp/pulp_python.conf
+
+RUN <<EOF
+if [[ "$PULP_UI_URL" ]]; then
+    mkdir -p /usr/share/nginx/static/pulp_ui
```
mdellweg: Is this the new /var/www ?
It's OK...

images/pulp/stable/Containerfile
```diff
@@ -12,7 +12,9 @@ ARG PULP_MAVEN_VERSION=""
 ARG PULP_PYTHON_VERSION=""
 ARG PULP_RPM_VERSION=""
 ARG PULP_OSTREE_VERSION=""
+ARG PULP_UI_URL=""
 
+ENV PULP_UI=true
```
mdellweg: This should probably depend on PULP_UI_URL... Or let's just make PULP_UI_URL the env var?

images/s6_assets/nginx.conf.j2
```diff
@@ -105,6 +105,17 @@ http {
 
         include pulp/*.conf;
 
+        {% if ui | default(false) -%}
```
mdellweg, line 108: This could be `if ui_url != ""`?

mdellweg (commented):
I like it a lot.

images/pulp/stable/Containerfile
gerrod3: Where should these files live? I was thinking I could put them in the pulp static directory (default: /var/lib/operator/static/), but I was worried if the user ran collectstatic with --clean then they would lose their UI as well.

images/pulp/stable/Containerfile
```diff
@@ -39,3 +41,10 @@ USER root:root
 RUN ln $(pip3 show pulp_ansible | sed -n -e 's/Location: //p')/pulp_ansible/app/webserver_snippets/nginx.conf /etc/nginx/pulp/pulp_ansible.conf
 RUN ln $(pip3 show pulp_container | sed -n -e 's/Location: //p')/pulp_container/app/webserver_snippets/nginx.conf /etc/nginx/pulp/pulp_container.conf
 RUN ln $(pip3 show pulp_python | sed -n -e 's/Location: //p')/pulp_python/app/webserver_snippets/nginx.conf /etc/nginx/pulp/pulp_python.conf
+
+RUN <<EOF
+  if [[ "$PULP_UI_URL" ]]; then
+    mkdir -p "${PULP_STATIC_ROOT}pulp_ui"
+    curl -Ls $PULP_UI_URL | tar -xzv -C "${PULP_STATIC_ROOT}pulp_ui"
+  fi
+EOF
```
mdellweg: I haven't seen this "here document" thing in a container file before.
Can you try with `\` on each lines end instead?

gerrod3: I couldn't get it to work with `\`.

gerrod3: Ok I got a version with `\` to work, had to switch to using single `[` for the if check.

ggainey (approved):
This all looks reasonable now, nice work.

mdellweg (approved):

PR #687, opened 2024-10-21
Delete base-images cache entry on rebuild

gerrod3:
So now that we are using the cache between different runs to check if we need to rebuild the base-images, we need to delete the existing entry when a rebuild occurs since the cache action doesn't replace the existing entry upon reupload.


mdellweg (approved):
Oh crap...

Hard problem number one.

PR #694, opened 2024-11-20
Make rebuilding the app images smarter

gerrod3:
When running the build app image CI for a release (nightly, on-merge, manual) the build action will check to see if it can skip rebuilding the image if no new versions of Pulp have been released for that branch. This should make our nightly runs do less work and avoid publishing what are practically duplicate images.

### DIFF
--- a/.github/actions/build_image/action.yml
+++ b/.github/actions/build_image/action.yml
@@ -16,6 +16,9 @@ inputs:
     description: 'Use the latest pulp-ui when building the image'
     default: 'false'
     required: false
+  built_base_images:
+    description: 'A JSON list of the base-images that were freshly rebuilt prior'
+    required: true
 outputs:
   app_version:
     value: ${{ steps.image_version_branch.outputs.app_version }}
@@ -23,6 +26,9 @@ outputs:
   app_branch:
     value: ${{ steps.image_version_branch.outputs.app_branch }}
     description: 'The pulpcore version branch that the built image matches'
+  rebuilt_images:
+    value: ${{ env.BUILD }}
+    description: 'true/false if the app image was rebuilt'
 
 runs:
   using: "composite"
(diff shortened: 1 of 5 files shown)


.github/actions/build_image/action.yml
```diff
@@ -62,7 +68,42 @@ runs:
...
+        # 2. Base images were rebuilt
+        # 3. New pulp versions was released
+        if [[ "${{ github.event_name }}" == "pull_request" || "${{ inputs.image_variant }}" == "nightly" || -n "${{ inputs.built_base_images }}" ]]; then
+          echo "BUILD=true" >> $GITHUB_ENV
+        else
+          if [[ "${{ steps.cache.outputs.cache-hit  }}" == "false" ]]; then
+            echo "BUILD=true" >> $GITHUB_ENV
+          else
+            # Script returns non-zero (100) when new versions are available
+            if python .ci/scripts/check_for_updates.py ${{ github.ref_name }} versions.freeze; then
```
mdellweg: This name would help reading the if clause:
```suggestion
            if python .ci/scripts/check_up_to_date.py ${{ github.ref_name }} versions.freeze; then
```

.github/actions/build_image/action.yml
```diff
@@ -84,11 +125,26 @@ runs:
...
+      run: |
+        echo "Deleting existing cache for ${{ env.VERSIONKEY }}"
+        gh cache delete "base-images=${{ env.VERSIONKEY }}" -R ${{ github.repository }}
+      shell: bash
+
+    - name: Cache versions
+      if: env.BUILD == 'true' && github.event_name != 'pull_request'
+      uses: actions/cache/save@v4
+      with:
+        key: ${{ env.VERSIONKEY }}
```
mdellweg, line 149: Is this the same cache that we key with `base_images={{ env.VERSIONKEY }}`?

mdellweg (approved):
I did not know that you can get dependencies from pypi without fetching the actual wheels. This is nice!

.github/actions/build_image/action.yml
gerrod3: It's the same cache container, but not the same entry. `base_images={{ env.VERSIONKEY }}` was a typo left over from copy-pasting. I fixed it, now the key should only be `{{ env.VERSIONKEY }}` for the `versions.freeze` file.

mdellweg (approved):

PR #695, opened 2024-11-21
Fix find_pulpcore_version script for new pyproject.toml changes

gerrod3:
Switch to reading the pyproject.toml to find the current version for main.


mdellweg (approved):
It's always the main branch, so I guess we don't need a fallback path...

PR #700, opened 2024-12-09
More fixes to app-rebuild logic

gerrod3:
Second round of fixes:

1. The output, `rebuilt_images` didn't seem to be getting set, so switched it to how all the other outputs are set.
2. Add some more debugging statements so I can see what is being stored in the cache
3. When we don't do a rebuild there is no image to get versions from, so return back the cached version

### DIFF
--- a/.github/actions/build_image/action.yml
+++ b/.github/actions/build_image/action.yml
@@ -27,7 +27,7 @@ outputs:
     value: ${{ steps.image_version_branch.outputs.app_branch }}
     description: 'The pulpcore version branch that the built image matches'
   rebuilt_images:
-    value: ${{ env.BUILD }}
+    value: ${{ steps.rebuild_needed.outputs.build }}
     description: 'true/false if the app image was rebuilt'
 
 runs:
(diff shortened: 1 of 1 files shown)


.github/actions/build_image/action.yml
```diff
@@ -121,9 +124,13 @@ runs:
       id: image_version_branch
       run: |
         base_image=$(echo ${{ inputs.image_name }} | cut -d '-' -f1)
-        app_version=$(podman run --pull=never pulp/${{ inputs.image_name }}:ci-amd64 bash -c "pip3 show pulpcore | sed -n -e 's/Version: //p'")
+        if [[ "${{ env.BUILD }}" == "true" ]]; then
```
mdellweg: Should we not rely on the env here?
```suggestion
        if [[ "${{ steps.rebuild_needed.outputs.build }}" == "true" ]]; then
```

gerrod3: Yep, remove `env.BUILD` since it's no longer needed.

mdellweg (approved):

PR #730, opened 2025-03-12
Properly download the correct s6 overlay per machine arch

gerrod3:
The last change was making it so that the ARM s6 files were the last ones extracted which is why the container wasn't working no more (at least when running on x86).

### DIFF
--- a/images/pulp_ci_centos/Containerfile
+++ b/images/pulp_ci_centos/Containerfile
@@ -2,8 +2,7 @@ ARG FROM_TAG="latest"
 FROM pulp/base:${FROM_TAG}
 # https://ryandaniels.ca/blog/docker-dockerfile-arg-from-arg-trouble/
 
-RUN curl -Ls https://github.com/just-containers/s6-overlay/releases/download/v3.1.6.2/s6-overlay-x86_64.tar.xz | tar xvJ -C /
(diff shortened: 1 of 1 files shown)


ggainey (approved):
/me cheers wildly

images/pulp_ci_centos/Containerfile
```diff
@@ -2,8 +2,7 @@ ARG FROM_TAG="latest"
 FROM pulp/base:${FROM_TAG}
 # https://ryandaniels.ca/blog/docker-dockerfile-arg-from-arg-trouble/
 
-RUN curl -Ls https://github.com/just-containers/s6-overlay/releases/download/v3.1.6.2/s6-overlay-x86_64.tar.xz | tar xvJ -C /
-RUN curl -Ls https://github.com/just-containers/s6-overlay/releases/download/v3.1.6.2/s6-overlay-aarch64.tar.xz | tar xvJ -C /
+RUN curl -Ls "https://github.com/just-containers/s6-overlay/releases/download/v3.1.6.2/s6-overlay-$(arch).tar.xz" | tar xvJ -C /
```
mdellweg, line 5: Right. I just read "download a file" but this actually put's stuff into the container image.
Nice find!

PR #737, opened 2025-04-16
Add a CI job to test with pulp-operator

git-hyagi:


.github/workflows/ci.yml
```diff
@@ -108,3 +109,87 @@ jobs:
...
+      shell: bash
+    - name: Try the cluster !
+      run: kubectl get pods -A
+    - name: Setup a minikube docker env
+      run: minikube -p minikube docker-env | grep "export" | sed 's/export //' | sed 's/"//g' >> $GITHUB_ENV
+    - name: Build pulp-minimal and pulp-web images
+      run: |
+        docker build -f images/Containerfile.core.base -t pulp/base:latest .
+        docker build -f images/pulp-minimal/nightly/Containerfile.core -t localhost/pulp/pulp-minimal:ci .
+        docker build -f images/pulp-minimal/nightly/Containerfile.webserver -t localhost/pulp/pulp-web:ci .
```
gerrod3: Is it wise to test with nightly instead of stable?

.github/workflows/ci.yml
```diff
@@ -1,6 +1,7 @@
 name: pulp-oci-images CI
 
 on:
+  workflow_dispatch:
```
gerrod3: Don't think we need to run the CI on dispatch.

.github/workflows/ci.yml
```diff
@@ -108,3 +109,87 @@ jobs:
           sudo ls -al $VOLUME_PATH
           sudo tree $VOLUME_PATH
           http --follow --timeout 30 --check-status --pretty format --print hb http://localhost:8080/pulp/api/v3/status/ || true
+
+  pulp-operator:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - name: Set up Python 3.8
```
gerrod3: Can we use a later python like the rest of the workflow? The images run 3.8, but the machines running them don't have to.

gerrod3 (changes requested):
Can you also space out steps with newlines?

.github/workflows/ci.yml
git-hyagi: The idea of testing with nightly was to verify any possible incompatibility or error "earlier" (before a release). Do you think this is not necessary?

mdellweg: The question to ask is "What does it tell us?".
It is possible to produce false negatives?

git-hyagi: > The question to ask is "What does it tell us?".

hmm... something like, the current change in the image may break or be incompatible with pulp-operator?!

> It is possible to produce false negatives?

probably. I'm updating the PR with the stable image.

.github/workflows/ci.yml
gerrod3:
```suggestion
    - name: Set up Python 3.11
```

.github/workflows/ci.yml
```diff
@@ -108,3 +108,100 @@ jobs:
...
+        sleep 10 # giving some time for the pod to start
+        kubectl logs -f -l app.kubernetes.io/component=operator -c manager &
+        kubectl wait --for condition=Pulp-Operator-Finished-Execution pulp/example-pulp --timeout=900s
+      shell: bash
+
+    - name: Test all components
+      run: |
+        export INGRESS_TYPE=nodeport
+        export CI_TEST=true
+        .ci/scripts/pulp_tests.sh -m
```
gerrod3, line 202: Did you forget to push the test script file?

gerrod3 (changes requested):

gerrod3 (approved):

PR #739, opened 2025-04-18
Correct Pulp Web service dependencies in folder compose

Safranil:
The file is not in sync with compose.yml for the pulp_web service.

If wanted, I can merge this change with #738.

### DIFF
--- a/images/compose/compose.folders.yml
+++ b/images/compose/compose.folders.yml
@@ -73,7 +73,9 @@ services:
     command: ['/usr/bin/nginx.sh']
     depends_on:
       pulp_api:
+        condition: service_healthy
       pulp_content:
+        condition: service_healthy
     ports:
       - "8080:8080"
     hostname: pulp


images/compose/compose.folders.yml
```diff
@@ -73,7 +73,9 @@ services:
     command: ['/usr/bin/nginx.sh']
     depends_on:
       pulp_api:
+        condition: service_healthy
       pulp_content:
+        condition: service_healthy
```
mdellweg, line 78: I start to wonder if we need these dependencies at all.
I believe, nginx would just start fine and issue 502 until the backends are ready. Right?

Safranil: Indeed, this may not be necessary, but if the service is not started, the name is not recognized in Docker internal DNS. Can this lead to a 500 error instead of 502?

I will try and tell you.

Safranil: It's worse, Nginx can't start if the name aren't known, the service restart indefinitely:
```
pulp_web-1  | Nameserver is: 127.0.0.11 
pulp_web-1  | Generating nginx config
pulp_web-1  | Modifying /etc/nginx/pulp/pulp_ansible.conf
pulp_web-1  | Modifying /etc/nginx/pulp/pulp_container.conf
pulp_web-1  | Modifying /etc/nginx/pulp/pulp_python.conf
pulp_web-1  | Starting nginx
pulp_web-1  | 2025/04/22 11:43:11 [emerg] 1#1: host not found in upstream "pulp_content" in /etc/nginx/nginx.conf:43
pulp_web-1  | nginx: [emerg] host not found in upstream "pulp_content" in /etc/nginx/nginx.conf:43
```

So it's a requirement...

mdellweg: Thanks for the experiment!

Safranil: You're welcome 😄

mdellweg (approved):

mdellweg:
The docs workflow is broken on an old ci image... Unrelated to this change.

gerrod3 (approved):

PR #741, opened 2025-05-30
[PULP-445] Python 3.11

mdellweg:
This is a spike to see what actually needs to be adjusted.


mdellweg:
The question that remains: How can we teach plugin CI to pick this or the old python tag?

ggainey (approved):
And off we go to the Brave new World of py3.11 :)

PR #746, opened 2025-06-23
Upgrade postgresql to version 16

gerrod3:
https://issues.redhat.com/browse/PULP-592

Wrote the upgrade part to be generic to handle upgrading from any previous version to the next PG version, so theoretically upgrading the PG version next time should be easier.

### DIFF
--- a/images/pulp_ci_centos/Containerfile
+++ b/images/pulp_ci_centos/Containerfile
@@ -20,13 +20,29 @@ ENV S6_BEHAVIOUR_IF_STAGE2_FAILS=1
 # https://github.com/just-containers/s6-overlay/issues/467
 ENV S6_CMD_WAIT_FOR_SERVICES_MAXTIME=0
 
-RUN dnf -y install postgresql \
-      postgresql-contrib \
-      postgresql-server \
-      postgresql-upgrade \
+RUN dnf -y install "https://download.postgresql.org/pub/repos/yum/reporpms/EL-9-$(arch)/pgdg-redhat-repo-latest.noarch.rpm" && \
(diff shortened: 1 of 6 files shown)


images/pulp_ci_centos/Containerfile
```diff
@@ -20,14 +20,30 @@ ENV S6_BEHAVIOUR_IF_STAGE2_FAILS=1
 # https://github.com/just-containers/s6-overlay/issues/467
 ENV S6_CMD_WAIT_FOR_SERVICES_MAXTIME=0
 
-RUN dnf -y install postgresql \
-      postgresql-contrib \
-      postgresql-server \
-      postgresql-upgrade \
+RUN dnf -y install "https://download.postgresql.org/pub/repos/yum/reporpms/EL-9-$(arch)/pgdg-redhat-repo-latest.noarch.rpm" && \
```
decko, line 23: What's the difference for the packages from that repo to this one https://centos.pkgs.org/9-stream/centos-appstream-x86_64/postgresql-server-16.8-1.module_el9+1209+bd6e4013.x86_64.rpm.html ???

gerrod3: It's been a month since I wrote this, but if I recall correctly these modules can only be used 1 at a time but you need to be able to install the previous version alongside the new version to perform the upgrade. The modules solve this by providing the `postgresql-upgrade` which includes the previous version, but that wouldn't work for our case of jumping from 13 -> 16.

decko (approved):

mdellweg (approved):
I would not say that upgrading the database version is really in the scope of a ci container. But that we do not need to decide today.

PR #745, opened 2025-06-13
Enable nginx listening to IPv6 if available

mayrstefan:
Issue #595 was closed to early.

This is the required nginx template change to use IPv4 or IPV6 listeners depending on the system capabilities.

Attention: this PR requires a pulpcore version including the netutil.has_ipv6 function from https://github.com/pulp/pulpcore/pull/6594

PR #749, opened 2025-07-22
Revert "Enable nginx listening to IPv6 if available"

gerrod3:
Reverts pulp/pulp-oci-images#745
Currently breaking all of our other CI, reverting till we figure out why.


gerrod3:
Not needed.

PR #754, opened 2025-07-30
Remove unused packages in the base image

gerrod3:
This should trim 600 MB (install size) off the images, most of it from removing installing rust (cargo was just hanging out in our install list). The python system packages shouldn't be necessary, pip can install them, also they will be useless once we upgrade to 3.11. 

After this the next biggest packages are gcc, podman, buildah. gcc is needed to build three python dependencies that are only available as sdists: `rhsm`, `python-ldap` (from django-auth-ldap), and `PyGObject `(from pulp-ostree). The cairo & gobject rpms are also needed to build PyGObject. If we could somehow get these packages to wheels then we could shave off another 150 MB installed size.


mdellweg (approved):
This look absolutely great.

PR #755, opened 2025-08-05
Use native arm runner for github CI

gerrod3:
This should speed up our CI by removing the emulation we were doing before.

### DIFF
--- a/.github/actions/base_images/action.yml
+++ b/.github/actions/base_images/action.yml
@@ -1,15 +1,16 @@
 ---
 name: Build Base Images
 description: Build the base images (pulp/base & pulp/pulp-ci-centos9) if needed
-# Both ARM64 & x86-64 versions of each are built
-# Use hashFiles(base_image_files, pulp-ci_image_files) as the key to the cache
+# Use hashFiles(base_image_files, pulp-ci_image_files) for rebuild check
+inputs:
+  image_variants:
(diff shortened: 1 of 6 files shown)


gerrod3:
An overview of the changes:

Switching the CI runners to be by architecture: amd64 or arm64. Now all the images will be built one after another in the same runner as each build takes around 2-3 minutes now that we are not using emulation for the arm images. 

The base and build image actions have been updated to this new behavior. Since building is cheap now I removed the GH caching logic we were using for the base-images. Instead we now check if we need to rebuild and if not we pull the latest pushed build to use instead. This should eliminate a lot of redundant pushes we do where no significant changes happened.

 The biggest changes are in the publish workflow which now has to account for the two separate runners building the images. We need to reunite the amd & arm images in order to publish them, so we do a temporary upload to ghcr.io and then have a separate job pull the uploads, retag them and do the final manifest push if needed.
 
 Side note: the temporary upload might be a good candidate for using the Pulp service registry. The reason I didn't use the GH cache was because the total of all the images was too large and we could hit the cache limit.

dralley:
I didn't thoroughly review this (my shell scripting skills are terrible anyway) but this looks reasonable from a high-level view.  Removing emulation is hugely beneficial to performance, I did that recently on the createrepo_c / libcomps / libsolv build CI

ggainey (approved):
Approved - let's see what happens!

PR #759, opened 2025-08-27
Add container_build volume to installation docs

quba42:
There is an extra volume defined in the image itself, that is not documented anywhere.

Why is it needed? Should it be added to the installation instructions?

I don't understand the background, but I stumbled over it in usage.

I am hoping we can resolve the why on this PR, and then resolve it in such a way that the next person following the quick start guide does not stumble over this.

### DIFF
--- a/docs/admin/tutorials/quickstart.md
+++ b/docs/admin/tutorials/quickstart.md
@@ -28,17 +28,18 @@ This deployment is a 2-step process:
 1st, create the directories for storage/configuration, and create the `settings.py` file:
 
 ```
-$ mkdir -p settings/certs pulp_storage pgsql containers
+$ mkdir -p settings/certs pulp_storage pgsql containers container_build
 $ echo "CONTENT_ORIGIN='http://$(hostname):8080'" >> settings/settings.py
 ```
 
 * For a complete list of available settings for `settings.py`, see
   [the Pulpcore Settings](site:pulpcore/docs/admin/reference/settings/).
 
-* These 4 directories `settings`, `pulp_storage`, `pgsql`, `containers` must be preserved. `settings`
-  has your settings, generated certificates, and generated database encrypted fields key. The
-  `pulp_storage pgsql containers` are the application data.
-
+* The 3 directories `settings`, `pulp_storage`, `pgsql`, must be preserved. `settings` has your settings,
+  generated certificates, and generated database encrypted fields key. The `pulp_storage`, and `pgsql` store
+  persistent application data.
+* The `containers`, and `container_build` directories are expected to be used for temporary `pulp_container`
+  application data. We recommend preserving these as well, but they are less critical.
 
 ### Starting the Container
 
@@ -52,6 +53,7 @@ $ podman run --detach \
              --volume "$(pwd)/pulp_storage":/var/lib/pulp:Z \
              --volume "$(pwd)/pgsql":/var/lib/pgsql:Z \
              --volume "$(pwd)/containers":/var/lib/containers:Z \
+             --volume "$(pwd)/container_build":/var/lib/pulp/.local/share/containers:Z \
              --device /dev/fuse \
              pulp/pulp
 ```
@@ -66,6 +68,7 @@ $ podman run --detach \
              --volume "$(pwd)/pulp_storage":/var/lib/pulp \
              --volume "$(pwd)/pgsql":/var/lib/pgsql \
              --volume "$(pwd)/containers":/var/lib/containers \
+             --volume "$(pwd)/container_build":/var/lib/pulp/.local/share/containers:Z \
              --device /dev/fuse \
              pulp/pulp
 ```


docs/admin/tutorials/quickstart.md
```diff
@@ -52,6 +52,7 @@ $ podman run --detach \
              --volume "$(pwd)/pulp_storage":/var/lib/pulp:Z \
              --volume "$(pwd)/pgsql":/var/lib/pgsql:Z \
              --volume "$(pwd)/containers":/var/lib/containers:Z \
+             --volume "$(pwd)/???":/var/lib/pulp/.local/share/containers:Z \
```
quba42:
```suggestion
             --volume "$(pwd)/local_containers":/var/lib/pulp/.local/share/containers:Z \
```
Would this naming make sense?

goarsna: According to @gerrod3's comment in the [Pulp matrix chat](https://matrix.to/#/!HWvLQmBGVPfJfTQBAu:matrix.org/$UbOwAsYHsySVOYFs3pHT8XkpIhREuSreB-mWvdwT8Nk?via=matrix.org&via=ctrl-c.liu.se&via=fedora.im) I would suggest `build_containers`.

quba42: Or maybe `container_builds`?

goarsna: Sounds good to me :)

docs/admin/tutorials/quickstart.md
```diff
@@ -52,6 +52,7 @@ $ podman run --detach \
              --volume "$(pwd)/pulp_storage":/var/lib/pulp:Z \
              --volume "$(pwd)/pgsql":/var/lib/pgsql:Z \
              --volume "$(pwd)/containers":/var/lib/containers:Z \
+             --volume "$(pwd)/container_build":/var/lib/pulp/.local/share/containers:Z \
```
mdellweg, line 56: You probably want to add it to line 31 too.
Does it need to be preserved (line 38)?

quba42: I am not sure how critical it is to preserve that one. My understanding is that it is used by pulp_container to build images. As such, it presumably only needs to be preserved if you want to re-use your existing layers during re-build? I presume the final images once built are stored by pulp_container as Pulp artifacts just like any other content type. But I am guessing based on high level assumptions here.

quba42: For now, I added the extra folder in both places.

docs/admin/tutorials/quickstart.md
```diff
@@ -28,14 +28,14 @@ This deployment is a 2-step process:
...
 $ echo "CONTENT_ORIGIN='http://$(hostname):8080'" >> settings/settings.py
 ```
 
 * For a complete list of available settings for `settings.py`, see
   [the Pulpcore Settings](site:pulpcore/docs/admin/reference/settings/).
 
-* These 4 directories `settings`, `pulp_storage`, `pgsql`, `containers` must be preserved. `settings`
+* These 5 directories `settings`, `pulp_storage`, `pgsql`, `containers`, and `container_build` must be preserved. `settings`
   has your settings, generated certificates, and generated database encrypted fields key. The
   `pulp_storage pgsql containers` are the application data.
```
quba42: What should I say about the `container_build` folder here?

docs/admin/tutorials/quickstart.md
mdellweg: That'd be my understanding too. Thanks.

docs/admin/tutorials/quickstart.md
gerrod3: You can say it's a directory used by pulp_container to build images. Do we remember what the `containers` folder is used for? Could we also just put the `container_build` folder into the same folder?

gerrod3 (changes requested):
Sorry, just got back from PTO. See my comment about the existing `containers` folder.

docs/admin/tutorials/quickstart.md
quba42: `containers` is mounted to `/var/lib/containers` inside the container, so I guess your question translates to: What does Pulp store in `/var/lib/containers`?

quba42: This guide has something interesting to say on the matter: https://pulpproject.org/pulp-oci-images/docs/admin/guides/migrate-pulp-installer-to-multi-process/?h=%2Fvar%2Flib%2Fcontainers#create-an-empty-containers-directory

Quote:

> We do not bother to move /var/lib/containers because it is only ever used for temporary files by pulp_container, and may be used on the host for other purposes (like running podman)

quba42:
@gerrod3 How do you feel about this latest formulation? I mainly base my assertion that `container` and `container_build` contain "less critical data" on what I read here: https://pulpproject.org/pulp-oci-images/docs/admin/guides/migrate-pulp-installer-to-multi-process/?h=%2Fvar%2Flib%2Fcontainers#create-an-empty-containers-directory

If we are not confident in that statement I will just need to gloss over it somehow since I am working from "dangerous half-knowledge" here. :stuck_out_tongue_winking_eye:

gerrod3 (approved):
Seems reasonable to me. I'll need to spend some time and investigate if these are "truly" required or just a simple hack to make pulp-container happy. There might be a better way where we can drop needing to create these volumes.

PR #762, opened 2025-10-23
Fix nightly release flow and missing publish

gerrod3:
Couple issues to fix here:

1. The newly built images were not getting published. The publish action takes the arm & amd images and packages them together into a manifest list to push to the registries. When grabbing the two images we used `container-storage:localhost/pulp/{image}:ci-{arch}`, since when you build an image it gets auto tagged with the `localhost/` prefix. However, now that the publish step is in a different job and we are pulling down the images and retagging them the `localhost/` doesn't get applied so it seems the `podman manifest create` fails to find them.
2. The nightly release logic has been updated to reflect the new build logic: nightly never rebuilds the base images, it always uses latest, but it will always builds the app images. So the logic now accounts for this distinction when preparing to pass the images to the publish step.
3. Nightly can now fail without blocking the publishing of the stable images. Now that publish is a separate job it is reliant on the build-and-test job completing successfully, each of the four variants, before executing. We don't want nightly failing to build to stop us from publishing stable if it completely successfully. 
4. More logging to make sure behavior is easier to reason through.

### DIFF
--- a/.github/actions/base_images/action.yml
+++ b/.github/actions/base_images/action.yml
@@ -37,6 +37,7 @@ runs:
           config=$(oras manifest fetch-config --platform linux/amd64 ghcr.io/pulp/pulp-ci-centos9:${branch} | jq -r '.config.Labels')
           latest_hash=$(echo "${config}" | jq -r '."org.pulp.basefiles-hash"')
           build=true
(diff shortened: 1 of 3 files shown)


ggainey (approved):
Looks good. I really appreciate the extra logging too - will make it easier to diagnose should anything ever go worng again :)

PR #763, opened 2025-10-23
Undo publish change, fix nightly logic

gerrod3:
Ok, the latest run gave some interesting info: https://github.com/pulp/pulp-oci-images/actions/runs/18761033143/job/53527402366#step:5:425

My assumption about the `localhost/` prefix was wrong. It's still there even if we pull and retag, so I'm undoing that change. Not sure why the original run that didn't publish happened though (https://github.com/pulp/pulp-oci-images/actions/runs/18727980553/job/53423774152#step:6:97). It's possible this revert will make it still fail to publish while reporting that it did. A successful publish has more logs showing the blobs being uploaded: https://github.com/pulp/pulp-oci-images/actions/runs/18702581200/job/53335207305#step:6:99

And again I looked over the nightly logic and reworked it once more to be located to the area that is special, choosing the correct base images to pull.

### DIFF
--- a/.github/actions/publish_images/action.yml
+++ b/.github/actions/publish_images/action.yml
@@ -52,7 +52,7 @@ runs:
       run: |
         for image in ${{ inputs.image_names }}; do
           echo "Create manifest for ${image}"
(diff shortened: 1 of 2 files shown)


ggainey (approved):
Approving - assuming all goes green, we need to merge and then See What Happens

PR #765, opened 2025-10-24
Fix wrong tags for nightly base image

gerrod3:
Another simple fix. The good thing is last nights run showed that the nightly continue on error is working correctly and the stable images were published! The second purpose of this is to test the rebuild logic. The base images should only rebuild when there is a change in the base containerfiles or they have a package that needs updating, so going to watch their build step here to see that it gets skipped and uses last night's build. The app images always gets rebuilt on PRs, but once merge it will only get built when there is a new pulpcore or plugin version, so we should see it not get built there.


gerrod3:
Nevermind, the base images had updates so not able to test the rebuild logic. :cry:

ggainey (approved):

PR #773, opened 2025-11-19
Fix tag used for pulling when skipping an unbuilt image

gerrod3:
The current issue in the latest workflow is a missed optimization scenario, here's what's happening:

1. The latest version is 3.93.1, so latest, 3.93 and 3.93.1 tags point to the same manifest
2. The base image checks to see if they need to be rebuilt from the 3.93 tag, the answer is no.
3. The pulp image checks rebuild from 3.93 tag, answer is yes. New version is 3.93.2
4. The temp upload stage pushes the new pulp image just fine, but sets the tag to pull the base image from to 3.93.2
5. The publish step tries to pull the base image at 3.93.2 and errors since it doesn't exist!

Solution: When you don't rebuild set the unbuilt image to the branch of the workflow since that is what they checked against to see if they needed to build.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -178,31 +178,26 @@ jobs:
 
       - name: Temp upload images to ghcr.io
         run: |
-          trimmed_tag="${TAG::-4}"
+          branch=${{ github.base_ref || github.ref_name }}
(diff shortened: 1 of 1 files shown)


ggainey (approved):
Great problem-determination, this change looks to address it - approving

PR #781, opened 2026-01-12
Add check when upgrading from pg13

gerrod3:
Let's see if this works...

Assisted by: Cursor Composer-1


ggainey (approved):
This looks reasonable, and the baseurl is correct - approving.

PR #787, opened 2026-02-06
fix: include UI in stable ref builds, fix conditions

### LINKED ISSUE #712: Add pulp-ui to the pulp-minimal (web) image
Now that the UI is a part of the single container, we should also offer it as a part of the pulp-minimal + web containers. Since the UI is just static files it should probably be served from the web image.

gianklug:
This PR resolves #712.

Additionally, it fixes the conditional (`latest_ui`) for including the UI on `:stable` tags.

Currently, it is only set to `true` if `github.base_ref == 'latest'`, however `base_ref` is only set on PRs:
> `github.base_ref`: The `base_ref` or target branch of the pull request in a workflow run. This property is only available when the event that triggers a workflow run is either `pull_request` or `pull_request_target`. ([source](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#github-context))

We're now also allowing `github.ref_name == 'latest'`, so it also works for pushes to the `stable` branch.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -123,7 +123,7 @@ jobs:
         with:
           image_name: "pulp-minimal"
           image_variant: ${{ matrix.image_variant }}
-          latest_ui: ${{ github.base_ref == 'latest' }}
+          latest_ui: ${{ github.base_ref == 'latest' || github.ref_name == 'latest' }}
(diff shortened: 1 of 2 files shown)


gerrod3 (approved):
Thanks for the fix!

PR #794, opened 2026-03-02
Fix building images with non-supported plugin versions

### LINKED ISSUE #793: multiprocess image 3.73 uses unsupported pulpcontainer branch
**Version**
`docker.io/pulp/pulp:3.73.26`

**Describe the bug**

The 3.73 images uses the supported pulpcore 3.73.26 (good) but the unsupported pulpcontainer 2.25.1 (bad).
As a result, even when running an allegedly supported 3.73 pulp image, I don't get updates to pulpcontainer.

Here are the individual plugin versions from the image LABELs:
```
 LABEL
"org.pulp.plugins"="pulp-ansible==0.27.1
pulp-certguard==1.8.0
pulp-container==2.25.1
pulp-deb==3.5.3
pulp-file==1.16.0
pulp-gem==0.7.5
pulp-glue==0.32.3
pulp-glue-gem==0.5.3
pulp-maven==0.10.1
pulp-npm==0.2.0
pulp-ostree==2.4.8
pulp-python==3.17.1
pulp-rpm==3.29.10
pulpcore==3.73.26"
"org.pulp.postgres-version"="13.23"
"org.pulp.python-version"="3.9.25"
"org.pulp.containerfiles-hash"="d6fafe347d88b652ad663cb4e6433d2c173bd968a64d2ae18e588f4e20e4739c"
```

**To Reproduce**
Run an image from the 3.73 branch.

**Expected behavior**
Any of [pulpcontainer supported branches](https://github.com/pulp/pulp_container/blob/main/template_config.yml#L105) is used. Eg pulpcontainer 2.26.

**Additional context**

See https://github.com/pulp/pulp_container/pull/2201 which wil not be backported to the 2.25 pulpcontainer branch, therefore will never appear in the 3.73 branch of the pulp image.

gerrod3:
fixes: #793
Assisted by: claude-opus-4.6

### DIFF
--- a/.ci/scripts/check_up_to_date.py
+++ b/.ci/scripts/check_up_to_date.py
@@ -10,6 +10,7 @@ PACKAGES = [
     "pulp-container",
     "pulp-deb",
     "pulp-gem",
+    "pulp-hugging-face",
     "pulp-maven",
     "pulp-npm",
     "pulp-ostree",
@@ -24,11 +25,15 @@ PYTHON_VERSIONS = {
     Requirement("pulpcore>=3.85"): "3.11"
 }
 
-def check_update(branch, current_versions, should_exit=True):
+def check_update(branch, current_versions, plugin_specifiers=None, should_exit=True):
     """
     Go through each of the image's main Pulp components and see if there is a new version available.
     """
     new_versions = {}
+    plugin_specifiers = plugin_specifiers or {}
+    if plugin_specifiers and branch == "latest":
+        plugin_specifiers = {}  # latest branch should always check plugins' latest versions
+
     # Get the latest Z (or Y) pulpcore release for this branch
     core_pypi_response = requests.get(urljoin(INDEX, "pypi/pulpcore/json"))
     assert core_pypi_response.status_code == 200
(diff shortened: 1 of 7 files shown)


.ci/scripts/check_up_to_date.py
```diff
@@ -52,11 +57,18 @@ def check_update(branch, current_versions, should_exit=True):
...
+                    pass
+                else:
+                    break
             version_pypi_response = requests.get(urljoin(INDEX, f"pypi/{plugin}/{version}/json"))
             assert version_pypi_response.status_code == 200
             version_python_response_json = version_pypi_response.json()
             deps = version_python_response_json["info"]["requires_dist"]
+            if deps is None:
+                print(f"No requires_dist for {plugin} {version}")
+                continue
```
pedro-psb, line 71: Can this ever be None?

.ci/scripts/get_supported_specifiers.py
```diff
@@ -0,0 +1,95 @@
...
+    supported = set(tuples)
+    lo_maj, lo_min = tuples[0]
+    hi_maj, hi_min = tuples[-1]
+
+    parts = [f">={lo_maj}.{lo_min}.0"]
+    for minor in range(lo_min + 1, hi_min):
+        if (lo_maj, minor) not in supported:
+            parts.append(f"!={lo_maj}.{minor}.*")
+    parts.append(f"<{hi_maj}.{hi_min + 1}.0")
+    return ",".join(parts)
```
pedro-psb, line 72: 👍

pedro-psb (approved):
Look good to me

.ci/scripts/get_supported_specifiers.py
pedro-psb: You could use packaging, but it's fine...

.ci/scripts/check_up_to_date.py
gerrod3: Yes, apparently pulp_maven has a version where it is None. I don't think it is suppose to be None, should be an empty list, but had to add this check to prevent it breaking.

PR #803, opened 2026-03-10
Fix get_supported_specifiers returning constraints that were un-installable

gerrod3:
On some older branches there are no supported branches that can be resolved for that pulpcore version. Skip adding the constraint if so.

### DIFF
--- a/.ci/scripts/get_supported_specifiers.py
+++ b/.ci/scripts/get_supported_specifiers.py
@@ -5,10 +5,16 @@ and generate pip version specifiers that constrain installs to those branches.
 Can be imported as a module or run as CLI to output pip constraints format.
 """
 
+import argparse
 import requests
 from yaml import safe_load
+from packaging.requirements import Requirement
+from packaging.specifiers import SpecifierSet
+from packaging.version import parse
+from urllib.parse import urljoin
 
 GITHUB_RAW = "https://raw.githubusercontent.com/pulp/{repo}/main/template_config.yml"
+INDEX = "https://pypi.org"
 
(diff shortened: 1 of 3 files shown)


.ci/scripts/get_supported_specifiers.py
```diff
@@ -84,12 +137,16 @@ def get_specifiers(plugin_names=None):
         branches = get_supported_branches(repo)
         if branches:
             spec = build_specifier(branches)
-            if spec:
+            if spec and check_installable(pulpcore_version, pip_name, spec):
```
pedro-psb, line 140: 👍

.ci/scripts/get_supported_specifiers.py
```diff
@@ -72,8 +83,50 @@ def build_specifier(branches):
     return ",".join(parts)
 
 
-def get_specifiers(plugin_names=None):
+def check_installable(pulpcore_version, plugin, specifier):
+    """Check if the plugin and its constraints are installable with the given pulpcore version."""
+    for core_spread, core_python in PYTHON_VERSIONS.items():
+        if pulpcore_version in core_spread.specifier:
+            break
```
pedro-psb, line 90: Every time I read this for loop I think something like: "wait, that isn't doing anything...". Then I keep reading and realize the reason 😅

pedro-psb (approved):

.ci/scripts/get_supported_specifiers.py
gerrod3: Yeah it's a pretty ugly pattern, not sure if there is a more Pythonic way.

PR #816, opened 2026-04-30
Update deprecated GH actions

jobselko:


jobselko:
Note: `oras-project/setup-oras@v2` must be allowed at the org level

ggainey (approved):

PR #828, opened 2026-07-02
[3.73] Upgrade base images to Python 3.11 and PostgreSQL 16

gerrod3:
## Summary
- Upgrade base images from Python 3.9 to Python 3.11 with update-alternatives
- Upgrade embedded PostgreSQL from distro packages to PostgreSQL 16 via PGDG
- Replace postgres-init upgrade logic with pg_upgrade flow (including PG13 EOL repo support)
- Update pip3 references and hardcoded python3.9 site-packages paths
- Update CI constraint scripts to reflect Python 3.11 on these images

This backports the runtime upgrades from 3.85 to fix createrepo_c build/runtime
failures on Python 3.9 release branches.

## Test plan
- [ ] CI build-and-test-images passes on amd64 and arm
- [ ] pulp-operator job passes (3.73+)
- [ ] Fresh container starts and passes API status check
- [ ] Existing PG13/14 data directory upgrades cleanly to PG16 on container restart

Made with [Cursor](https://cursor.com)

### DIFF
--- a/images/s6_assets/init/postgres-init
+++ b/images/s6_assets/init/postgres-init
@@ -16,6 +16,7 @@ export C037="\e[37m"
 PGVERSION=$(postgres --version | egrep -o "1[0-9]" | head -n1)
 PGHOME="/var/lib/pgsql"
 PGDATA="${PGHOME}/data"
+PGDATA_NEW="${PGHOME}/data_new"
 PGDATA_OLD="${PGHOME}/data_old"
 PGDATA_OLD_RENAMED="${PGHOME}/data_old.$(date --rfc-3339=seconds)"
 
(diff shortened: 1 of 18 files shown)


images/s6_assets/init/postgres-init
```diff
@@ -24,23 +25,70 @@ if [ -d "${PGDATA}/base" -a -f "${PGDATA}/PG_VERSION" ]; then
...
+name=PostgreSQL 13 EOL for RHEL 9 - ${ARCH}
+baseurl=https://yum-archive.postgresql.org/13/redhat/rhel-9-${ARCH}/
+enabled=1
+gpgcheck=1
+gpgkey=https://yum.postgresql.org/keys/PGDG-RPM-GPG-KEY-RHEL
+EOF
+      echo -e "${PREFIX} ${ORANGE}Created repository file ${REPO_FILE}${ENDCOLOR}"
+    fi
+    dnf -y install "postgresql${PGDATA_VERSION}" "postgresql${PGDATA_VERSION}-server" "postgresql${PGDATA_VERSION}-contrib"
     # We have to check the encoding and the locale because pulp/pulp-oci-images #307 (1da694c) means it may differ.
```
dralley, line 52: So, we're installing the old postgresql exclusively to check the encoding and locale that were used previously?  There's no other way to do it?

How will it work across different base OS versions of the image?

dralley (approved):

images/s6_assets/init/postgres-init
gerrod3: You also need the old-binary to do the upgrade itself, see how the upgrade command takes in the new and old binaries. The encoding and locale check is to make sure that the new version uses the same values.

PR #834, opened 2026-08-06
Switch Pulp Nightly & debuginfod images to CentOS Stream 10

dralley:


dralley:
Sidenote: Should `pulp-minimal`'s nightly image be based off of the nightly pulp image?

https://github.com/pulp/pulp-oci-images/blob/latest/images/pulp-minimal/nightly/Containerfile.core#L1

The "webserver" variant is, I find it strange that this is not.

gerrod3:
> Sidenote: Should `pulp-minimal`'s nightly image be based off of the nightly pulp image?
> 
> https://github.com/pulp/pulp-oci-images/blob/latest/images/pulp-minimal/nightly/Containerfile.core#L1
> 
> The "webserver" variant is, I find it strange that this is not.

Not exactly. `pulp-minimal` inherits from the base image `pulp/base`. The nightly variant of `pulp-minimal` inherits from latest since there is no nightly variant of base. If we inherited from the `pulp/pulp` or `pulp/ci-centos` images we would end up with the s6 and postgres files in the `pulp-minimal` image, which we don't want. The `pulp-web` image inherits from the `pulp-minimal` image for one step in order to get the nginx snippets.  `pulp-web` is actually inheriting from nginx: https://github.com/pulp/pulp-oci-images/blob/latest/images/pulp-minimal/nightly/Containerfile.webserver#L11

gerrod3 (approved):

dralley:
In 1-2 weeks, as long as the nightly image doesn't surface any issues, I will switch over the base / minimal images.

PR #841, opened 2026-08-11
Increase nginx header buffer size for PQC certificates

dralley:
Post-quantum (ML-DSA) X.509 certificates are significantly larger than traditional RSA/ECDSA certificates. When a reverse proxy forwards a PQC client certificate via the X-CLIENT-CERT header, it can exceed nginx's default large_client_header_buffers size (4x8k), causing a 400 error before the request reaches the content app. Increase to 4x16k.

Assisted-By: Claude Opus 4.6

### DIFF
--- a/images/s6_assets/nginx.conf.j2
+++ b/images/s6_assets/nginx.conf.j2
@@ -20,6 +20,10 @@ http {
     # to build optimal hash types.
     types_hash_max_size 4096;
 
+    # PQC (post-quantum) X.509 certificates can exceed the default 8k buffer
+    # when forwarded as HTTP headers (e.g. X-CLIENT-CERT).
+    large_client_header_buffers 4 16k;
+
     upstream pulp-content {
          server {{ '[::1]' if has_ipv6 else '127.0.0.1' }}:24816;
     }


images/s6_assets/nginx.conf.j2
```diff
@@ -20,6 +20,10 @@ http {
     # to build optimal hash types.
     types_hash_max_size 4096;
 
+    # PQC (post-quantum) X.509 certificates can exceed the default 8k buffer
+    # when forwarded as HTTP headers (e.g. X-CLIENT-CERT).
+    large_client_header_buffers 4 16k;
```
dralley, line 25: Per Claude:

> The two arguments mean:
> 
> * 4 — the maximum number of buffers allocated per connection for reading large client request headers
> * 16k — the size of each buffer
> nginx reads each request header line into a buffer. By default this is 4 8k (4 buffers of 8KB each). A single header line (e.g. X-CLIENT-CERT: <url-encoded PEM>) must fit entirely within one buffer — it can't span multiple buffers. The multiple buffers are for handling multiple large header lines in the same request, not for splitting one line across buffers.
> 
> So 4 16k means: allow up to 4 header lines of up to 16KB each. The ML-DSA-87 cert is ~10.7KB URL-encoded, which exceeds the default 8KB single-buffer size but fits comfortably in 16KB.

The nginx docs for this field are here: https://nginx.org/en/docs/http/ngx_http_core_module.html#large_client_header_buffers

gerrod3 (approved):

PR #842, opened 2026-08-13
Fix several CI issues

jobselko:


.github/actions/test_image/action.yml
```diff
@@ -19,6 +19,11 @@ runs:
         pip install httpie podman-compose==1.1.0
       shell: bash
 
+      # Without slirp4netns, curl fails with "Connection reset by peer" on GHA ubuntu-24.04 20260810 runners
```
dralley: Do we have some idea why this happens, that we can document?

jobselko: No, this is the only thing I know. Something is wrong with podman 5.8.4 on the latest runner.

jobselko: I do not want to guess the reason based on the claude's hallucinations.

images/s6_assets/test.sh
```diff
@@ -18,7 +18,9 @@ trap cleanup EXIT
 # buildah & skopeo (pulp_container does)
 
 start_container_and_wait() {
+  # Without slirp4netns, curl fails with "Connection reset by peer" on GHA ubuntu-24.04 20260810 runners
   podman run --detach \
+             --network slirp4netns:port_handler=rootlesskit \
              --publish 8080:$port \
```
gerrod3:
```suggestion
             --publish 127.0.0.1:8080:$port \
```

Can you try this? Might also need to change the localhost:8080 down below as well.

jobselko: It is working, thanks!

gerrod3 (approved):

.github/workflows/ci.yml
```diff
@@ -153,9 +153,22 @@ jobs:
           echo "INFO:
           Updating registries configuration
           "
-          echo "[registries.insecure]
-          registries = ['ingress.local','nodeport.local:30000']
-          " | sudo tee -a /etc/containers/registries.conf
+          # registries.conf v1 and v2 formats cannot be mixed in the same file;
+          # detect which format the runner uses and append matching syntax.
+          if grep -q "unqualified-search-registries" /etc/containers/registries.conf; then
```
dralley, line 158: Is there going to be a time that we can drop compatibility with both versions?  Is this a Centos Stream 9 / 10 thing that could be simplified when we drop 9?

jobselko: I think this is a "GH runner images" thing

dralley: I thought all of our runners would be using the same OS though?

gerrod3: I believe this is more of a which podman version is the GH runner using vs the centos version inside the container. We need to do an audit of our supported branches and which versions of ubuntu they are using.

jobselko: The v1/v2 issue appeared in pulp_container, pulp-operator and here. I think it was when a new "Ubuntu 24.04 20260810.x" runner was used, while "Ubuntu 24.04 20260720.x" was working.

dralley: That is frustrating, but it would make sense.  Maybe we can switch to Ubuntu 26 soon (believe it's still in preview)

PR #843, opened 2026-08-14
Switch fully to CentOS Stream 10 for official non-CI images

dralley:


gerrod3 (changes requested):
If we are going to make the full switch I would like to remove the old ci-centos9 image and rename the base-cs10 as just base. The older branches will still publish the old images, but every new image should be cs10 based.

dralley:
> would like to remove the old ci-centos9 image

We can't do that, the older branches of pulpcore / plugins are still using the old images, and will be for a while.  I don't think we want to switch old branches over to the new image.

gerrod3:
> > would like to remove the old ci-centos9 image
> 
> We can't do that, the older branches of pulpcore / plugins are still using the old images, and will be for a while. I don't think we want to switch old branches over to the new image.

Yes we can. The older branches are still going to publish centos9, it's just the latest branch will not.

dralley:
How far do you want to go with it?  Because ultimately we're going to have to do this over again in the future.

We could ditch the suffixes on anything public-facing, and keep using them for e.g. `pulp-ci-centos10`.

gerrod3:
On the latest branch I would like to just publish `base`, `pulp-ci-centos10`, `pulp` and `pulp-minimal/web`. Only the CI image should have the suffix, the others should be indifferent to it. Users will do an upgrade and now be running on a newer OS, no changes needed. When we need to bump the OS again we can just repeat what we have done and create a separate temporary image for testing and initial switching, then replace the current with new once we are sure everything is good.

gerrod3 (approved):
Nice!

PR #846, opened 2026-08-19
[PR #842/40a90890 backport][3.73] Fix several CI issues

gerrod3:
**This is a backport of PR #842 as merged into latest (40a9089054fe0e9aa4ae40afb94d9513f0d0235a).**

Cherry-pick conflict: 3.73 does not have the `setup-go@v6` step that was already present on `latest`, so that step was not added.

## Summary
- Use `--pull=missing` when building the pulp-web image so the local pulp-minimal:ci is reused
- Append insecure registries in the format the runner actually uses
- Workaround pulp-operator test script until pulp-operator#1667 lands
- Bind the test container to 127.0.0.1:8080

## Test plan
- [ ] CI build-and-test-images passes
- [ ] pulp-operator job passes

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -156,9 +156,22 @@ jobs:
           echo "INFO:
           Updating registries configuration
           "
-          echo "[registries.insecure]
-          registries = ['ingress.local','nodeport.local:30000']
-          " | sudo tee -a /etc/containers/registries.conf
+          # registries.conf v1 and v2 formats cannot be mixed in the same file;
+          # detect which format the runner uses and append matching syntax.
+          if grep -q "unqualified-search-registries" /etc/containers/registries.conf; then
+            echo '
+          [[registry]]
+          location = "ingress.local"
+          insecure = true
+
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -186,6 +199,12 @@ jobs:
       with:
         repository: pulp/pulp-operator
 
+    # Remove once https://github.com/pulp/pulp-operator/pull/1667 is merged
+    - name: Fix push repo type in operator test scripts
+      run: |
+        sed -i 's/repository -t push show/repository show/g' .ci/scripts/pulp_container/export_repository.sh
+      shell: bash
+
```
jobselko: This PR uses an older pulp_container version, so this should not be here.

gerrod3: Removed. These branches ship an older pulp_container that still uses `repository -t push show`, so that operator test-script workaround from latest does not apply.

jobselko (approved):

PR #852, opened 2026-09-08
Fix image build failing to remove a missing file

mdellweg:
Can't we just boldly stop trying to remove that file?


mdellweg:
I think the issue revealed now is unrelated...

gerrod3 (approved):

